PR #155, opened 2015-01-06
Add South Africa as a core country supported by MapIt

mhl:
There's some interest in an independent MapIt being set up for South Africa,
rather than just the private one mounted in the People's Assembly site.

This pull request adds support for South Africa as a core country, in the same
way as Norway, the UK or Global.

For the moment, there's an instance you can try out here:

* http://ec2-54-220-26-94.eu-west-1.compute.amazonaws.com/

### DIFF
--- a/mapit/context_processors.py
+++ b/mapit/context_processors.py
@@ -2,7 +2,10 @@ from django.conf import settings
 
 
 def country(request):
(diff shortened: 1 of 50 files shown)


dracos:
Some minor comments, otherwise :golf:

PR #156, opened 2015-01-06
Use "CREATE EXTENSION postgis; ..." on Ubuntu trusty

mhl:
As described here:

  https://docs.djangoproject.com/en/1.6/ref/contrib/gis/install/postgis/#creating-a-spatial-database-with-postgis-2-0-and-postgresql-9-1

... creating a PostGIS database should be done with
"CREATE EXTENSION postgis; CREATE EXTENSION postgis_topology;"
where PostGIS and PostgreSQL are late enough versions, rather
than the script to create a template_postgis template.

The only supported distribution where we have these versions is
currently Ubuntu trusty.

### DIFF
--- a/bin/install-as-user
+++ b/bin/install-as-user
@@ -68,7 +68,19 @@ fi
 echo -n "Setting up database... "
 if ! psql -l | egrep "^ *$DB_NAME *\|" > /dev/null
 then
-    createdb -T template_postgis --owner "$UNIX_USER" "$DB_NAME"
+    if [ trusty = "$(lsb_release -s -c)" ]
+    then
+        # Currently the only supported distribution where we can just
+        # use CREATE EXTENSION to make this a PostGIS database is on
+        # Ubuntu trusty:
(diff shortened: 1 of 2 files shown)


dracos:
I guess this could future-proof a bit and put debian jessie in already? That looks to have 9.4/2.1.

dracos:
Also, looking at our `.travis.yml` it appears that Django 1.4 does not support using the postgis extension for working with geo database - the commands there work for installing on recent versions of PostgreSQL/PostGIS. This I don't think matters as the install script will use the system Django which is 1.6.1 on trusty, but not sure if this is worth noting/ dealing with in some way.

stevenday:
Not sure if it will affect this or not, but FYI, it looks like Django 1.8 is going to call `CREATE EXTENSION postgis` automatically: https://docs.djangoproject.com/en/dev/releases/1.8/#django-contrib-gis

dracos:
Django uses `IF NOT EXISTS` so it won't matter if it already exists.

PR #164, opened 2015-01-28
Stream JSON responses

dracos:
When we start, we have an iterable `QuerySet`. So this changes `add_codes` to be iterable, passes an `iterdict` [1] rather than a `dict` to `output_json`, changes `output_json` to use a `StreamingHttpResponse` by default, and updates the callback middleware to work with streaming or non-streaming responses. This hopefully reduces memory usage for one particular request from c. 2Gb to 200Mb. Similarly for `output_html`, we keep everything as an iterable, doing something horrible with a template in order to put our iteration in the middle.

[1] An `iterdict` is a `dict` subclass that does nothing but provide an iterable for `items`/`iteritems`, in order so that the `json` package will think it's a dictionary, iterate over it and output an iterator. Otherwise we'd need to load everything into a dictionary in memory to iterate it all back out...

### DIFF
--- a/mapit/shortcuts.py
+++ b/mapit/shortcuts.py
@@ -40,18 +40,12 @@ def render(request, template_name, context=None):
     )
 
 
-def sorted_areas(areas):
-    # In here to prevent a circular dependency
-    from mapit import countries
-    if hasattr(countries, 'sorted_areas'):
-        return countries.sorted_areas(areas)
-    return list(areas)
 
 
 def output_html(request, title, areas, **kwargs):
     kwargs['json_url'] = request.get_full_path().replace('.html', '')
     kwargs['title'] = title
-    kwargs['areas'] = sorted_areas(areas)
+    kwargs['areas'] = areas
     kwargs['indent_areas'] = kwargs.get('indent_areas', False)
     return render(request, 'mapit/data.html', kwargs)
 
--- a/mapit/shortcuts.py
+++ b/mapit/shortcuts.py
@@ -1,13 +1,18 @@
+import itertools
 import json
+
 import django
 from django import http
 from django.db import connection
 from django.conf import settings
 from django.shortcuts import get_object_or_404 as orig_get_object_or_404
 from django.shortcuts import render_to_response
-from django.template import RequestContext
+from django.template import loader, RequestContext, Context
 from django.utils.six.moves import map
 from django.utils.encoding import smart_str
+from django.utils.translation import ugettext as _
+
+from mapit.iterables import defaultiter
 
 from django.core.serializers.json import DjangoJSONEncoder
 # Assuming at least python 2.6, in Django < 1.6, the above class is either a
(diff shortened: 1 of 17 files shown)


dracos:
Note that:
* this doesn't deal with the corresponding HTML response; Django templates are not streamable, that's ticket https://code.djangoproject.com/ticket/13910 – to work around this we could work on that ticket, and/or construct a dummy template and insert our sequence for streaming;
* although streaming responses have been around since Django 1.5, it seems no-one has tried putting Unicode in them and gzipping them – https://code.djangoproject.com/ticket/24240;
* or indeed tried gzipping them before at all – https://code.djangoproject.com/ticket/24242.
Hohum. I have worked around 24240 with 9943537, 24242 we could fork our own copy of the gzip middleware or monkey-patch.

mapit/shortcuts.py
```diff
@@ -33,47 +40,64 @@ def default(self, o):
...
-    if hasattr(countries, 'sorted_areas'):
-        return countries.sorted_areas(areas)
-    return list(areas)
-
-
 def output_html(request, title, areas, **kwargs):
     kwargs['json_url'] = request.get_full_path().replace('.html', '')
     kwargs['title'] = title
-    kwargs['areas'] = sorted_areas(areas)
     kwargs['indent_areas'] = kwargs.get('indent_areas', False)
```
davea: Wasn't this the line you decided to chop?

mapit/shortcuts.py
```diff
@@ -33,47 +40,64 @@ def default(self, o):
...
-    return render(request, 'mapit/data.html', kwargs)
+    tpl = loader.render_to_string('mapit/data.html', kwargs, context_instance=RequestContext(request))
+    wraps = tpl.split('!!!DATA!!!')
+
+    def template_area(area):
+        return loader.render_to_string('mapit/areas_item.html', {'area': area, 'indent_areas': kwargs['indent_areas']})
+    areas = map(template_area, areas)
```
davea: This could probably be sped up by not loading/compiling the template for each area (even if template caching is being used, it'd save a round-trip to the cache). Something like this maybe?

```
    item_tpl = loader.get_template('mapit/areas_item.html')
    areas = map(lambda area: item_tpl.render(Context({'area': area})), areas)
```

mapit/shortcuts.py
dracos: Yeah, but then I need to pass it in the `template_area` iterator so thought I might as well leave it for that.

davea:
Aside from those little inline comments, :+1:

mapit/shortcuts.py
dracos: Of course, dur, thank you.

mapit/shortcuts.py
dracos: But it doesn't need to be in kwargs so changed for the below.

PR #165, opened 2015-02-11
Add django 1.8 support

dracos:
And tidy up bits for some future proofing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ install:
   - sudo apt-get update -qq
   - sudo apt-get install -qq gdal-bin binutils
   - pip install $MODULES
-  - pip install -r requirements.txt
+  - CFLAGS="-O0" pip install -e .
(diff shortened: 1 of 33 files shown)


mhl:
Incidentally, why are you preferring `get_version()` (returning a string) to `VERSION` (being a tuple)? It seems like that latter ought to be more robust (although presumably it'll be a while until Django 1.10, so I guess it's not an immediate concern - just curious.)

mhl:
I've added a few comments or questions, but otherwise this looks good to me.

dracos:
"why are you preferring get_version()" – `get_version` is all that's mentioned in the documentation, so I assumed was preferable, it's a PEP386 compliant string, and Django will be going from 1.9 to 2.0 so should be okay for a while :)

dracos:
Merged in fc7f2eb8cf4ef76783a83cc96bbc12b62633de44 with Version 1.3 commit added on top separately to fix typos.

PR #167, opened 2015-02-26
I18n it

guglielmo:
There are a couple of unwanted changes (blank lines added or missing). 

Then I changed the way ``mapit_it`` urls are appended to the main urls, since ``patterns`` is not imported. 
Import instructions for ``render`` was wrong in ``mapit_it/urls.py``.
Finally a ``mapit_it`` template was missing the ``{% load i18n %}`` statement.

### DIFF
--- a/mapit/management/find_parents.py
+++ b/mapit/management/find_parents.py
@@ -27,6 +27,9 @@ class FindParentsCommand(NoArgsCommand):
             generation_low__lte=new_generation, generation_high__gte=new_generation,
         ):
             parent = None
+            self.stdout.write("Processing %s" % (
+                self.pp_area(area)
+            ))
             for polygon in area.polygons.all():
                 try:
                     args = {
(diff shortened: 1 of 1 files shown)


guglielmo:
I have no idea why it says I want to merge 14 commits. I really wanted to merge only the last one. But the diff in files seem to be ok.

dracos:
Hi – thanks for this :) "I have no idea why it says I want to merge 14 commits." – there appears to have been some confusing arising from merges and rebases. I have taken your 07ba1c5, 250122b, 9eb6150, 726d875 and f80f70b02 commits and rebased/merged them into our `i18n-it` branch okay, please do take a look.

I have not included aece5e335a349df09ee3637c5af1b179dd6f73f5 – I think these were correct already, assuming Italian also prints co-ordinate as (latitude,longitude)? And I changed "lookup by point" to match the other headings, which I hope is okay.

If you could check that our `i18n-it` works okay for you compared with your current branch; if so, then I suggest you `git reset --hard mysociety/i18n-it` on your `i18n-it` branch to bring it back in sync.

The only thing I would suggest before we can merge this is that you change the helpful "Processing" you've added to `find_parents.py` to only be output when the verbosity is set to 2 or more, so it's not shown by default. If you're happy to do that, we can then get this merged in :)

PR #170, opened 2015-03-03
management tasks improvements

guglielmo:


guglielmo:
6d64bea makes ``FindParentsCommand`` emits a logging message that it's processing an area, only when verbosity is set to 2 or more.

24bad64 does a few things:

* adds the ``mapit_export_geojson_areas`` task, to export multiple areas as a geojson ``FeatureCollection``.
* modifies ``mapit_import_area_unions, polishing it, generalizing it, and making it suitable to be uses in a general context
* adds the ``mapit_IT_generate_constituencies_csv`` task, specialized for the italian context, that automatically produces the CSV file needed by ``mapit_import_area_unions`` out of the DB.
* adds the ``mapit_IT_import_postal_codes`` task, also specialized for the italian context, that tries a naif strategy of importing postcodes, using OpenStreetMap's nominatim service (needs refining).

dracos:
Thanks for this :) For the postal code import script, was there a reason you didn't subclass the command, instead creating your own one? See how the UK one works at https://github.com/mysociety/mapit/blob/master/mapit_gb/management/commands/mapit_UK_import_codepoint.py – it looks like all you'd need to do is subclass the `handle_row` function and add nominatim to the `option_list`, everything else is identical.

Could you rebase this to put your fixing commits in with the previous commit?

guglielmo:
Thanks! There was no particular reason, it was just the first thing I did. I had to re-define the command options though, since I wanted ``--coord-field-lat`` to default to None, instead of 2.
I'm getting to know how to use rebase. You'll figure out the mess is due to this.

dracos:
Sorry for the delay; I have rebased and fixed the history issues in the mysociety `i18n-it` branch, splitting into three commits - improved area unions, geojson export, and the Italian management scripts. I have `flake8`d your changes, and tweaked the postal code importer to not have to redefine the command options but subclass the parser instead.
Do you want the CSV files included? As well as the two present, there was also cap_totale.csv which only appeared in the merge commit 8b1204ec91b69156d125cce01e27a4b126016fc6, so I wasn't sure what was happening there. Once I know what to do about them, I can merge this in.

dracos:
Reopening as mistakingly closed, though lack of feedback means this does need dealing with, eithier unilaterally merging or ignoring.

PR #171, opened 2015-03-03
Catch errors extracting names more selectively

### LINKED ISSUE #168: mapit_import reports it couldn't find a name field when errors occur reading the name field
This section: https://github.com/mysociety/mapit/blob/master/mapit/management/commands/mapit_import.py#L217-L230 of the `mapit_import` command tries to extract the name of a feature from the supplied data file, and prints a helpful error if it can't find a name under the field name you specified. Later on, it then tries to decode the name using the encoding you specified.

However, because the Django GDAL api uses `force_text` on the return value of fields, decoding errors can happen too. The first try/except catches them and so reports the wrong message to the user.

I'll take a stab at fixing this, this is just to remind me what the problem is.

stevenday:
Closes #168


stevenday:
Hmm, the test passes fine for me locally but fails on Travis. It's failing because it's not raising an error during the import at all when I'm trying to test the error message is the right one - I wonder if it's something about my versions? (Python 2.7.3, Django 1.7)

dracos:
I do get the same error locally, python 2.7.9, django 1.7.1. I think the difference will be in the version of GDAL (or underlying library) that is installed; it looks like my version is returning the string with the \x92 encoded as the illegal UTF-8 byte sequence \xc2\x92 which nonetheless `decode()`s in python fine.

mapit/management/commands/mapit_import.py
```diff
@@ -215,16 +215,15 @@ def verbose(*args):
...
             else:
                 try:
-                    name = feat[name_field].value
-                except:
-                    choices = ', '.join(layer.fields)
-                    raise CommandError(
-                        "Could not find name using name field '%s' - should it be something else? "
-                        "It will be one of these: %s. Specify which with --name_field" % (name_field, choices))
-                try:
+                    name = feat.get(name_field)
```
dracos: I think Feature's `get()` throws an `OGRIndexError` exception if name_field isn't present, unlike the normal pythonic `get()`, so the `None` check won't be reached.

stevenday:
So since my PR has it's own problems, and it's really a problem with the underlying GDAL library, I don't think this is much of an improvement, I'll retract!

PR #175, opened 2015-03-20
Add country-specific templates to MANIFEST.in

stevenday:
These were left out of the package by mistake, meaning you couldn't
use them from a different project when you installed django-mapit,
which is a shame.

This adds recursive-include statements for all of the current country apps.


dracos:
Thanks for spotting this! I've just, in a similar vein, spotted the country apps are missing the compatibility change of 83c2b82, so I've added that to this branch, hope that's okay :) Then this could be a quick 1.3.1 bugfix release before getting all the i18n stuff done for 1.4 :rabbit:

PR #176, opened 2015-03-23
Add template blocks for extra css and js into base.html

stevenday:
This will allow me to add some extra styling for the mapit gb api user
accounts, without having to fork the base template.


dracos:
:+1:

PR #177, opened 2015-03-23
Proposed version 1.3.1

stevenday:
I've updated the CHANGES.txt and bumped the version in setup.py, @dracos is
there anything else I need to do before I tag it and release to PyPi?

(Note: I'd like to merge #176 into this too)


dracos:
That should be all you need to do :)

PR #178, opened 2015-03-26
Contact tweaks

abimysoc:
The enquiries@ address gets quite a bit of what might be described as "tech support". This PR:

a) adds a contact address (mapit@) for technical queries

b) attempts to make there be only one place where the contact email addresses appear, so that changing them isn't quite so involved next time; hopefully I caught all the other places.


dracos:
Merged as 6945d8b

PR #179, opened 2015-03-30
Make sure CSS is compiled when running setup.py.

dracos:
Some unrelated flake8 fixes.

### DIFF
--- a/mapit/djangopatch.py
+++ b/mapit/djangopatch.py
@@ -6,6 +6,7 @@ if django.get_version() >= '1.8':
     from django.template.loader import render_to_string
 else:
     from django.template import loader, RequestContext
+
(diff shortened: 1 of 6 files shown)


stevenday:
Tested and working for me - I wonder though, if we can run checks, could we just run bin/mapit_make_css? 

I guess it assumes that the machine you're doing the packaging on is setup with sass etc, and that you'd want this to happen silently...

dracos:
I did wonder about that; I wasn't sure if you were in this situation whether you'd want to check that the CSS was working okay (perhaps it had come from elsewhere but it was you doing the packaging), or if the CSS compilation failed in some way. Not hard to run it if this ever does happen, I guess :)

PR #182, opened 2015-04-14
Add pricing info to the licensing page

### LINKED ISSUE #181: Add some pricing info to the licensing page

stevenday:
This is mostly copied from the new text on the TWFY page, along with some
re-arrangements to match the general tone of that such as moving the techy
stuff lower down.

@abimysoc, @MyfanwyNixon - could you check this over too? (I'll add a
screenshot in a second).

Closes #181

### DIFF
--- a/mapit/templates/mapit/licensing.html
+++ b/mapit/templates/mapit/licensing.html
@@ -9,49 +9,65 @@
 <p>Questions about how MapIt works, or the data? Drop a line to
 <a href="mailto:mapit&#64;mysociety.org">mapit&#64;mysociety.org</a>.</p>
 
-<h2>Usage and licensing</h2>
+<h2>Terms of usage</h2>
+<p>
+    Low volume, charitable use of the API is free. This means direct use by
+    registered charities, or individuals pursuing a non&ndash;profit project on an
+    unpaid basis, with a volume of up to 50,000 calls per year. All other use
+    requires a licence, as a contribution towards the costs of providing this
+    service. Please email us at
+    <a href="mailto:enquiries&#64;mysociety.org">enquiries&#64;mysociety.org</a>
+    if your usage is likely to fall into the non&ndash;free bracket:
+    <strong>unlicensed users may be blocked without warning</strong>.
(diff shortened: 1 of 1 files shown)


stevenday:
screen shot 2015-04-14 at 10 10 47

MyfanwyNixon:
All good - but I think we need to be more explicit about which of our 'helpful' services would be paid for. 
The following points need 'prices start at', 'for a cost of' or (if applicable) 'this service is free' attached to them IMO:

"We can set you up with a dedicated service"
"Some organisations prefer to install their own internal version of MapIt. We are happy to help with this. Depending on your needs, we can advise your technical team on how to do this, or our developers can install it for you"
One off data range services - "we can do this for you quickly and easily"

abimysoc:
IAWMyf.

"our developers can install it for you (for a fee)"
"we can do this for you - ask us for a quote" (I've removed "quickly and easily" here because pricing is based on time spent, not on difficulty.)
"We can set you up with a dedicated service" already mentions "paid" earlier in the paragraph so I think it's implied there that there'd be a cost; if someone is interested in lifting the rate limit they'd have to talk to us about buying a licence anyway.

Plus I would also request that the 'Technical queries' section is put back at the top of the page, where it is currently.

BenJam:
might be worth having a >500,000 option of 'get in touch'.

stevenday:
Sorry, realised that I never added a second screenshot - here's how it looks now:

screen shot 2015-04-17 at 11 26 49

abimysoc:
https://github.com/mysociety/theyworkforyou/pull/814#issuecomment-94020130 also applies here.

stevenday:
This is all rebased up and ready to go I think - @dracos, can you give it a look over before I merge and deploy?

dracos:
If I wanted to be really picky, given you've used nice typography elsewhere, I'd ask for an en-dash rather than a hyphen where you've used one ;) Looks fine.

stevenday:
@dracos, nice dashes everywhere as requested - ta :)

PR #184, opened 2015-05-09
Add Sweden to list of MapIt countries

jonkri:


dracos:
Hi :) A few random questions - you say postal codes are seven characters, but http://en.wikipedia.org/wiki/Postal_codes_in_Sweden says they are five?
And do you really want to allow anything through as a valid postcode? :)

I think there might be a bug in terms of partial postcodes, to do with length assumption, but we can look at that after some more details. I'm coincidentally just merging the i18n branch which means you can supply a Swedish translation too if you wish :)

jonkri:
Hi, Matthew!

I was not the one who wrote this code, I just found that it was never submitted back to the project, and made this pull request without even noticing that.

Swedish postcodes are for sure not seven digits. I will try to get hold of the person who wrote the code and ask, though. :) I'll get back to you.

(It's somewhat common that people write post codes by prepending "SE-" like so: "SE-41516". The other common format, except the "41516" one, is "415 16", with a space between the last two digits.)

jonkri:
Hi, again!

So, no one knows how that got in there. It would suffice for us to simply use "\d{3}\s?\d{2}", to match either "41516" or "415 16".

What would it mean to supply a Swedish translation? How many strings would that be to translate, approximately? Can we do it through Transifex?

Thanks!

dracos:
Do update this pull request then :) I'll take a look at the partial postcode issue soon.
Yes, the translation is on Transifex - https://www.transifex.com/projects/p/mapit/ - you don't need to translate the country specific resources, only 'mapit' itself, which is 166 strings. This would be so your installation of MapIt could be in Swedish, that's all :)

jonkri:
Sure, I will!

Could you create a "mapit_se" resource and add the Transifex account "jonkri" to it, please?

dracos:
Sorry, that probably isn't clear; you don't need a mapit_se resource as you have nothing country specific to translate – you only need to translate the "mapit" resource. I've invited you to it :)

jonkri:
I have now translated the resource to Swedish. :)

dracos:
Added to version 1.4 branch as 15e205ecb7d692917f3c9292902c6ce4f7364527, will be merged soon, thanks :)

PR #185, opened 2015-05-10
I18n for merging

dracos:


osfameron:
Looks ok (after quick scan, minor comments to matthew in person)

PR #187, opened 2015-05-12
Add a missing {% endif %} endblock tag

mhl:


dracos:
Thanks. Whoops! Test suite sets COUNTRY to GB otherwise it would have spotted it, I see :/

PR #191, opened 2015-05-16
Add fuzzy matching (issue #101)

### LINKED ISSUE #101: Enable fuzzy string searching/matching for areas
Searching by string prefixes is great but not nearly as useful as fuzzy string searching would be. Fuzzy string searching makes it possible to geocode with data that includes misspellings and variations.

Read this [question on GIS.SE](http://gis.stackexchange.com/q/79510/23321) for more background on this request.

chris48s:
See issue #101 for further info/notes on this feature.
Submitted for approval, but happy to discuss or do further work on this if required.
Fixes #101

### DIFF
--- a/mapit/postgres_trigram.py
+++ b/mapit/postgres_trigram.py
@@ -1,0 +1,25 @@
+"""
+Classes that represent PostgreSQL database functions.
+"""
+from django.db.models.expressions import Func, Value
+from django.db.models.fields import FloatField
+from django.contrib.postgres.lookups import PostgresSimpleLookup
+from django.db.models import CharField, TextField
+
+
+class TrigramBase(Func):
+    def __init__(self, expression, string, **extra):
+        if not hasattr(string, 'resolve_expression'):
+            string = Value(string)
+        super(TrigramBase, self).__init__(expression, string, output_field=FloatField(), **extra)
+
+
+class TrigramSimilarity(TrigramBase):
+    function = 'SIMILARITY'
+
+
+@CharField.register_lookup
+@TextField.register_lookup
+class TrigramSimilar(PostgresSimpleLookup):
+    lookup_name = 'trigram_similar'
+    operator = '%%'
--- a/mapit/views/areas.py
+++ b/mapit/views/areas.py
@@ -252,10 +252,25 @@ def areas_by_type(request, type, format='json'):
 
 @ratelimit(minutes=3, requests=100)
 def areas_by_name(request, name, format='json'):
+    fuzzy = request.GET.get('fuzzy', '')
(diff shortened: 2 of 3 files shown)


chris48s:
Hi.
I realise it has been some time since I followed up on this, but I recently submitted a minor patch to update djorm-ext-pgtrgm for Django 1.9 compatibility ( https://github.com/jleivaizq/djorm-ext-pgtrgm/pull/19 ) so it seemed like a good time to revisit this.

I've updated this PR so that /areas has a param ?fuzzy=true as suggested by @dracos in the discussion around issue #101 instead of using its own endpoint. I've also removed the code to pull this package in as a git submodule given this is not your preferred approach.
As you'll note from https://github.com/jleivaizq/djorm-ext-pgtrgm/pull/19 I have raised the issue of updating the package on pypi with the original author but got no response. To review the issue of how best to pull in this dependency using pip, it occurs to me that pip can install the package directly from GitHub using the command
```
pip install git+git://github.com/jleivaizq/djorm-ext-pgtrgm.git
```
or
```
pip install git+git://github.com/jleivaizq/djorm-ext-pgtrgm.git@6a5eefd
```
(if you want to 'lock' the dependency to a particular commit). Does this sound like a reasonable solution to this issue?

You'll note that the Travis CI build fails on this commit with the error 'ImportError: No module named djorm_pgtrgm' which is unsurprising given the above. I assume that this is not something I am able to address, given the dependencies for the test build are pulled in from
```
/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages
```

Finally, if there are any other issues relating to this feature (see issue #101 for previous discussions) that you would like me to do anything further on, let me know and I will follow up.

Thanks

dracos:
"I assume that this [Travis failure] is not something I am able to address" - Travis (well, tox) only installs what `setup.py` is telling it, which is what you'll get when you install mapit from pypi, so it does need to be fixed before it can be merged :) I'm afraid you can't install source control sources in `setup.py`, only actual packages, so (though see below) the only solution in the absence of upstream update would be for someone to create their own package on pypi containing this code and we can then reference that.

However, as we only support Django 1.8+ now, adding such a lookup is only a few lines, I don't think we need the third party module at all for that. This could be done similarly to the usage example I gave at https://github.com/django/django/pull/4815/files#diff-7c3d79c64cab66e031bd47e81345207fR29 with the lookup/functions given in that PR, which would be straightforward to include. The 'CREATE EXTENSION' (and I guess index) could be done with a migration, though I'm not sure if it shouldn't be left as a manual step in case someone doesn't have the extension installed. Does including the required lookups directly here sound okay to you?

chris48s:
That makes things easier then - thanks

Issue #74, opened 2013-07-23
possibly add an extra_types parameter to MapIt queries


mhl:
(This came up in a discussion in IRC of possibly adding NHS CCG areas to MapIt UK - there are [shapefiles available](http://www.england.nhs.uk/resources/ccg-maps/).)

One of the things that discourages us from adding new, potentially useful dataset to MapIt UK, is that the time for point-in-polygon lookups goes up when new datasets are added.  For smaller polygons this should not make much of a difference, but with large polygons (e.g. Police areas) @JennyD and I did some timings and found that the existence of those polygons in the database slowed down any unrestricted lookup by about 17%.

It's always possible to restrict the types of the search (with the `?type=`) query parameter, but @jacksonj04 suggested that one could also add a converse option: have some non-default types that could be searched.  For example, some types might not be searched by default, but would be if included in an `extra_types` parameter specified them.

I think this is something worth considering - my biggest concern would be that when people evaluate MapIt, I guess they typically do it by searching for a particular point, and seeing what areas are returned, so the `extra_types` areas wouldn't be so obviously discoverable.

symroe:
Adding to this – police force IDs is something that Democracy Club will need for the 2016 elections.

Given the performance hit that @mhl pointed out above I wonder:

1. If advances in PostGIS and general hardware speed in the last couple of years might mitigate the 17% slow down seen or

2. If DC should host it's own install of MaPit that only contains shapes for elections that we care about (i.e., not parishes, yet).

Given that we'll want this quick soon (on the timeline of this 2 year old issue), option 2 might be best for us.

PR #192, opened 2015-05-21
Add import script for NHS England CCG areas

chris48s:
As with the Police Force boundaries (see pull request #73), NHS CCG boundaries for England are available as open data, but they aren't formatted in such a way that they import nicely into MapIt.
This import script solves that problem, importing the CCG boundaries cleanly.
This may be useful to pull into master for users self-hosting, or could be useful for mapit uk if you decide to publish the CCG boundaries (as raised in issue #74).

### DIFF
--- a/mapit_gb/management/commands/mapit_UK_import_NHS_CCG_areas.py
+++ b/mapit_gb/management/commands/mapit_UK_import_NHS_CCG_areas.py
@@ -1,0 +1,204 @@
+# This script is used to import NHS England CCG area boundaries into MapIt.
+# These boundaries are available as open data from:
+#
+# http://www.england.nhs.uk/wp-content/uploads/2013/03/ccg-kml-format.zip
+#
+# Unfortunately, they aren't very usefully formatted - the names
+# are embedded in an (incomplete) HTML table in the description field. e.g:
+#
+# <description><![CDATA[<br><br><br>
+#    <table border="1" padding="0">
+#    <tr><td>CCGcode</td><td>00C</td></tr>
+#    <tr><td>CCGname</td><td>NHS Darlington CCG</td></tr>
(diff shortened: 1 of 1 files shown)


dracos:
I realise this copies the previous PR you mention, but is there a reason this command isn't a subclass of the mapit_import Command? I can't obviously see a reason why it has to use call_command and duplicate a lot of options/set up code, but I may well have missed something.

chris48s:
Commits have been tidied as requested. I'll have a go at refactoring this based on your suggestions and get back to you with an updated pull request.

chris48s:
I've refactored this to extend mapit_import, as requested. I've constructed an appropriate option_list by slicing mapit_import's option_list tuple. This solution is more succinct then redefining it but the downside to this is that adding or removing elements from the start or in the middle of mapit_import's option_list tuple would require updating the line:
```
    option_list = mapit_import.Command.option_list[:7] + \
                  mapit_import.Command.option_list[8:11] + \
                  mapit_import.Command.option_list[15:20]
```
accordingly.
Just something to be aware of there.

dracos:
Thanks, that's a bit less code :) A suggestion for your downside - see how what I've done in https://github.com/mysociety/mapit/commit/3175d648060e9f0c3c0f966a88c0fd86db43cc4b uses `create_parser` to set defaults instead, doing it that way (or other similar things, perhaps `remove_option`) might save you having to do a slice, and be more robust.

chris48s:
No problem - updated as requested.

torotil:
I think this is outdated as the current CCG boundary files can be imported with the generic import command.

chris48s:
Agreed :+1:  Closing this and tidying up some other forgotten PRs that went nowhere

PR #193, opened 2015-05-21
GeoJSON/KML for multiple areas (issue #136)

### LINKED ISSUE #136: Support GeoJSON for multiple areas
It would be great if the /areas/foo,bar,baz API supported returning GeoJSON, instead of just the json metadata.

Eg: http://global.mapit.mysociety.org/areas/509521,509535.geojson would be a FeatureCollection of http://global.mapit.mysociety.org/area/509521.geojson and http://global.mapit.mysociety.org/area/509535.geojson

chris48s:
The idea raised in issue #136 would be a really useful feature so I've had a stab at adding it.

A few notes on this:
1. In order to avoid repetition, I've factored out the KML strings in area.export into another class. I hope this is ok.
2. I've decided to only support KML and geojson formats. To add WKT support would require use of GEOMETRYCOLLECTION which is not widely supported (I actually had trouble locating a WKT parser that supports GEOMETRYCOLLECTION to test with) so I've left out WKT support. Happy to revisit this if others disagree with this.
3. Potentially this allows a user to create calls which would take a long time to execute by passing a large number of areas. If this is of concern with relation to usage limits, I could add a restriction on the number of areas that will be accepted.

Let me know if there is further work you would like to see on this.

Fixes #136.


dracos:
Thanks for also looking at this :) (You have some unrelated commits present here also.)

Factoring out the strings into another class is fine, though I think it could be done in a slightly more Pythonic way. From what you've written, it seems to me you want a serialiser class that is given an area or list of areas in its constructor, and does the KML/JSON construction internally, rather than relying on being passed in already generated KML and JSON snippets and gluing them together. `export()` could similarly use this serialiser, it's just on one object rather than a list. So in pseudocode, skipping any non-relevant bits/error handling:

```python
def areas_polygon(request, area_ids, srid='', format='kml'):
    ...
    area_ids = area_ids.split(',')
    areas = Area.objects.filter(id__in=area_ids)
    serialiser = GeometrySerialiser(areas, srid, simplify_tolerance)
    if format == 'kml':
        output = serialiser.kml(line, edge)
    elif format == 'json':
        output = serialiser.json()
```

`export()` could then use `serialiser = GeometrySerialiser([self], srid, simplify_tolerance)` etc. And the kml/json functions would both call the same function to get all the polygons for the passed-in areas, only differing in their output. Does that make sense?

Other notes: I see Django 1.8 has a GeoJSON serialiser - https://docs.djangoproject.com/en/1.8/ref/contrib/gis/serializers/ - though your code of course supports older versions and KML :) I guess performance is a consideration, though perhaps not necessarily just in number - asking for lots of tiny areas might be fine compared to a few huge ones. If this call would be caught by the database timeout other calls have, that would be fine, otherwise it might need to have its own timeout somehow (I think that's preferable to a hard limit on number of areas, though that might be needed too).

chris48s:
Commits have been tidied as requested. I'll have a go at refactoring this based on your suggestions and get back to you with an updated pull request.

chris48s:
Hi,
I've now refactored this based on your comments. Hopefully this is more in line with what you were looking for. I've also added a few unit tests - which I should have done in the first place.

Notes:
1. As you rightly say, the number of areas it is reasonable to request depends on the type of areas you are requesting. 10 LSOAs is probably completely reasonable whereas 10 European regions probably isn't. I did see if this could be controlled using set_timeout(), but this doesn't enforce a meaningful restriction. This is because the time-consuming bit of this request isn't the database query. Fetching areas by ID is a fast query - even with a lot of areas (as opposed to a spatial query like 'touches' or 'covers' where most of the work is done in the DB query). Given that, I'm not entirely sure how best to address this but I'm happy to revisit the topic if you have another suggestion.
2. I've retained the existing concatenation-based method of serialisation for GeoJSON as it would probably be useful to maintain a bit of backwards-compatibility on Django versions but if you're not concerned about that, let me know and I'll rewrite GeometrySerialiser.geojson() to use Django 1.8's serialiser.

Cheers

dracos:
Thanks :-) Looks good. Yes, I agree with your point 2. On your point 1, I guess we can just see how it goes. It might be nice to have the KML/JSON lazily evaluate, as happens with the normal list views (see my recent blog post at https://www.mysociety.org/2015/06/01/django-streaminghttpresponse-json-html/) but that can probably wait! Another independent PR adds a GeoJSON export script, which I will rewrite to use your serialiser as that gives KML export too.

longhotsummer:
I think `/areas/1.geojson` should always return a `FeatureCollection` -- that's what the plural `areas` implies. I've included that as a fix in https://github.com/chris48s/mapit/pull/1.

longhotsummer:
Any idea when you'll be able to look at this?

dracos:
Apologies, as the tests were marked as failing, I assumed work was still being done/awaited on this. If the tests could be fixed (looks to be only some small flake8 issues), then I can hopefully take a look early next week.

chris48s:
Sorry - my fault. Updated PR. Build now passing.

dracos:
Hi, Sorry for the delay in looking at this! I have rebased, made a few small alterations (e.g. factored some stuff shared between area_polygon and areas_polygon to common functions), and merged onto master. Thanks very much for your contribution :-)

PR #199, opened 2015-07-09
Include same details in .geojson as in .json

longhotsummer:
This includes the same details in an area's .geojson representation as in its .json representation. This is very useful when using .geojson on a webpage via CORS and saves you making two requests.

We have to build our own geojson manually, because the Django support for geojson doesn't gives us the opportunity to plug in custom properties.

### DIFF
--- a/mapit/models.py
+++ b/mapit/models.py
@@ -1,5 +1,6 @@
 import re
 import itertools
+import json
 
 from django.contrib.gis.db import models
 from django.contrib.gis.gdal import SRSException, OGRException
@@ -365,13 +366,29 @@ class Area(models.Model):
             else:
                 raise Exception("Unknown kml_type: '%s'" % (kml_type,))
             content_type = 'application/vnd.google-earth.kml+xml'
-        elif export_format in ('json', 'geojson'):
+        elif export_format == 'json':
             out = all_areas.json
+            content_type = 'application/json'
+        elif export_format == 'geojson':
+            out = json.dumps(self.as_geojson(polygons=all_areas, srid=srid))
             content_type = 'application/json'
         elif export_format == 'wkt':
             out = all_areas.wkt
             content_type = 'text/plain'
         return (out, content_type)
+
+    def as_geojson(self, polygons=None, srid=4326):
+        polygons = polygons or self.polygons.all().collect()
+        out = {
+            'type': 'Feature',
+            'crs': {
+                'type': 'name',
+                'properties': {'name': 'EPSG:%d' % srid},
+            },
+            'properties': self.as_dict(),
+            'geometry': json.loads(polygons.geojson),
+        }
+        return out
 
 
 @python_2_unicode_compatible
--- a/mapit/models.py
+++ b/mapit/models.py
@@ -381,12 +381,29 @@ class Area(models.Model):
         polygons = polygons or self.polygons.all().collect()
         out = {
             'type': 'Feature',
+            'properties': self.as_dict(),
+            'geometry': json.loads(polygons.geojson),
+        }
+
+        if srid:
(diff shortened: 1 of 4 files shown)


longhotsummer:
This also adds support for returning geojson for multiple areas at once.

dracos:
Thanks for this :) From a quick look, one issue is if we have users of the current .geojson endpoint who are assuming that it gives the geometry directly, they would have their usage break. Perhaps we could add a query parameter (to either .json, .geojson or both) that could signify you wanted the additional information? Not very nice, but something like ?with_geometry=1 on .json or ?with_data=1 on .geojson. Or perhaps you can think of a nicer solution, new endpoint entirely perhaps?

Bit of overlap/links with #193, just to check that everything merges nicely.

longhotsummer:
I didn't realise the slight discrepancy with the existing .geojson, that sucks :(

Perhaps a URL like `/area/12/feature.geojson` to preserve backwards compatibility -- what do you think?

I saw #193 after submitting this pull request. I'm happy to update this pull request to use #193 once that is merged, I like that it cleans up the export.

longhotsummer:
I've been looking at #193 and I'd MUCH rather build this on top of #193, which makes things much cleaner. Is there an ETA on when #193 will be merged?

longhotsummer:
I've submitted a pull request to include these changes to @chris48s branch, see https://github.com/chris48s/mapit/pull/1

chris48s:
Thanks @longhotsummer. On reflection you are right that it would be more consistent for /areas to always return a FeatureCollection.

@dracos What will be the easiest way for you to deal with this?
Is it best for me to pull @longhotsummer 's pull request into my fork so that pull request #193 then covers both 'GeoJSON/KML for multiple areas' and 'Include same details in .geojson as in .json'?
Or,
Will it be easier for you to review/merge this if #193 is updated such that /areas/1.geojson returns a FeatureCollection instead of returning the same output as /area/1.geojson but then keep 'Include same details in .geojson as in .json' as a seperate unit of functionality with its' own pull request?

Let me know and I will update accordingly.

dracos:
@chris48s Thanks for the detailed query – I think the latter, it would be preferable for #193 to be kept as multiple areas with the FeatureCollection fix added, then we can get that PR merged in, and then deal with this on top, with the change to use a different endpoint as discussed.

longhotsummer:
I think my pull request has muddied the waters. So, to clarify, https://github.com/chris48s/mapit/pull/1 now changes @chris48s's pull request to fix `/areas/1.geojson` to return a FeatureCollection. It does NOT change the behaviour of /area/1.geojson -- which is the primary concern with *this* pull request.

So it should be safe to merge #193  without breaking backwards compatibility.

Once that is done, I'll re-work this pull request completely, and we can add `/area/1/feature.geojson` with an improved properties bag.

Hope that makes it simpler :)

dracos:
#193 has been merged, so happy to go ahead with this whenever you're ready :)
If I've understood it right, it involves changing geometryserialiser's `geojson` to include more in properties; and adding a new URL that calls a tweaked area_polygon/export (perhaps making `kml_type` a general `type` so it could perform the same sort of task for both KML and GeoJSON.

longhotsummer:
Thanks @dracos, I'll take a look. I need to remind myself what's going on here :)

dracos:
I think this has been superseded by https://github.com/mysociety/mapit/pull/376 - this adds the dict information to the areas call, doesn't change the area call - I thought about adding a new endpoint as discussed above, but I think if someone wants this they can request /areas/ID.geojson instead of /area/ID.geojson and get a FeatureCollection with one Feature that contains the extra information.

PR #202, opened 2015-09-10
Don't allow through bad float numbers.

dracos:


stevenday:
Thinking out loud... this will match:
`1`
`-1`
`1.2`
`-1.2`

But it'll also match: `.2` and `-.2`, is that ok?

dracos:
```
>>> float(.2)
0.2
>>> float(-.2)
-0.2
```
Yep :) I thought it more likely someone would put `.2` then `2.` (which also would be fine to accept, but made it easier to have one `\d+` somewhere so it wasn't all optional).

stevenday:
cool! :+1:

PR #204, opened 2015-09-10
Add a command to raise the generation_high on current areas

mhl:
This is useful if you've created a new generation for a new data
import, but there are existing areas in the database that should
still remain active after you've activated your new generation.
After creating the new inactive generation you can run this command
to raise the generation_high of any areas who have a generation_high
set to the current (i.e. most recent active) generation so that
they're the newly created generation.

You could do this in the Django shell, but this seems like a safer
way of doing this.


dracos:
I wonder if it could have optional country/type arguments – quite often when updating OS data I wish to update a particular country and/or type (e.g. because Boundary-Line is GB only).

mhl:
@dracos I've added a `--country` and `--type` option to restrict the areas that are being modified to those matching the country code or area type code - does that look OK?

dracos:
Looks good! I'd squash them together :)

PR #209, opened 2015-10-27
Drop support for Django before version 1.8

dracos:
Slightly early, as 1.7 has extended support until December, but might as well get it out the way.


stevenday:
Wow, so much deleted code! This is great :rocket: 

FWIW, I tested this on a local vm that was running Django 1.7 before, and the one thing I had to do was clear my cache (as well as running migrations) - I wonder if it's worth adding some kind of upgrade note to this effect when you release this? (As well as bumping & signposting the version number so that people using older Djangos can find a version that works?)

stevenday:
Ah, one more thing, I just noticed we get a warning:

```
WARNINGS:
?: (1_8.W001) The standalone TEMPLATE_* settings were deprecated in Django 1.8 and the TEMPLATES dictionary takes precedence. You must put the values of the following settings into your default TEMPLATES dict: TEMPLATE_DEBUG.
```

PR #211, opened 2015-11-05
Fix GDAL error in mapit_import tests

stevenday:
When running the mapit tests as a group, something is leaving GDAL drivers
open/registered (I'm not sure of the terminology), but only Raster ones.

Django's Driver class checks for the existence of drivers before registering
new ones, but it assumes that if any exist, they all do: https://github.com/django/django/blob/b35adb0909b25a7dafc9212ddedfbf9b29dc05b8/django/contrib/gis/gdal/driver.py#L81-L83

In these tests that was not the case, so calls to the Vector driver resulted
in a pointer error. This commit provides a temporary workaround which force-
registers a driver (if none already exist) before the test runs.

### DIFF
--- a/mapit/tests/test_import_commands.py
+++ b/mapit/tests/test_import_commands.py
@@ -4,12 +4,20 @@ from django.test import TestCase
 from django.core.management import call_command
 from django.conf import settings
 from django.utils.six import StringIO
+from django.contrib.gis.gdal.prototypes import ds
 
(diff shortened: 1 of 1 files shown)


dracos:
I said okay to this already, I swear, but it's not here. Pity I can't replicate (something to do with loading the admin page causing the raster to be registered, but doesn't happen here), but this is fine.

PR #212, opened 2015-11-06
Fix JSONP issues

dracos:
An alternative solution to 2a30c9e would be for non-200 responses, consuming the generator immediately, as it'll always be small. Not sure if that's preferable/clearer.


stevenday:
:+1: I think 2a30c9e is as good as the alternative you suggest, since it can be used in all cases (200 or not).

dracos:
Merged at 66e8f90 (fixed a comment typo).

PR #217, opened 2015-11-27
October 2015 Boundary-Line

dracos:
Couple of issues to work around this time:
- [x] Incorrect parish IDs
- [x] Parish has come back
- [x] Wales & Scotland have been wrong in Boundary-Line for ages: "It transpires that the National Assembly for Wales constituencies and regions have been updated prematurely within Boundary-Line releases Oct 2013-May 2015, regarding the Parliamentary Constituencies and Assembly Electoral Regions (Wales) (Amendment) Order 2011. These will be amended to their correct alignment in the October 2015 release. Please note that this Order and the amendments do not come into force until the next National Assembly for Wales’s election in May 2016; they will be made available in the May 2016 Boundary-Line release."

### DIFF
--- a/mapit_gb/controls/2009-10.py
+++ b/mapit_gb/controls/2009-10.py
@@ -14,6 +14,6 @@ def code_version():
     return 'ons'
 
 
-def check(name, type, country, geometry):
+def check(name, type, country, geometry, **args):
     """Should return True if this area is NEW, False if we should match"""
     return False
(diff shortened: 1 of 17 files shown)


stevenday:
Minor comments aside :+1:

torotil:
This does not seem to be working for imports into completely new databases. Either `first-gss` also must include the fix for the incorrect parish IDs or control `2015-10` should be used with new databases - which also doesn't work because of the other "checks". How is this supposed to work?

dracos:
Ah, sorry, I should have put the patch in the main script, whoops. I've opened a new PR which hopefully means it will work as a first import.

torotil:
Thanks ;)

PR #225, opened 2016-02-16
Changes after a fresh import

symroe:
Here are 3 commits I made whilst importing the data from scratch.

`Patch incorrect IDs in first-gss control file` is because `patch_boundary_line` is needed on the initial import, however the patch function was missing from the `first-import` file.

`Open CSV files with the correct encoding` might just be an issue with python 3, but I think it's backwards compatible.

And the last one is just a small pep8 update that I thought should be it's own commit.

### DIFF
--- a/mapit_gb/controls/first-gss.py
+++ b/mapit_gb/controls/first-gss.py
@@ -14,3 +14,17 @@ def code_version():
 def check(name, type, country, geometry, **args):
     """Should return True if this area is NEW, False if we should match"""
     return False
+
+
+def patch_boundary_line(name, ons_code, type):
+    """Used to fix mistakes in Boundary-Line. This patch function should return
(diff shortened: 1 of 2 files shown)


dracos:
Patch IDs - ah, because if it's a first import you can't use the 2015-10 control file as is, I see. As these patches might only apply to the 2015-10 edition but will always be correct, it's okay to have them in the first use file. It would be good if this could be edited to state the edition of Boundary-Line that these patch lines exist for.

CSV - sure, happy to add this, but this is a very old one off script from 2010, the data it uses won't have been updated since then; just checking you realise that.

Break lines - We have a setup.cfg in the repository root that sets our max line length to 119, so this isn't necessary, thanks :)

dracos:
First fixed differently, second cherry picked, third unnecessary, so this is done, thanks :)

PR #226, opened 2016-02-16
Police import

symroe:
As a helper to others, this fixtures file creates the types needed to run the police areas import command with the following:

`./manage.py mapit_UK_import_police_force_areas --commit --generation_id 1 --area_type_code PDG --name_type_code "P" --code_type "police_id" [path/to/police/kmls]`

I don't know where to document this, as the main docs don't seem to include anthing about the police importer.


dracos:
Thanks - could the binary commit please explain why it's necessary?

PR #228, opened 2016-02-23
Use reverse so more redirects work when MapIt is mounted on a subpath

mhl:
If you've included MapIt in another Django project, e.g. with:

    urlpatterns += patterns('',
        (r'^mapit/', include('mapit.urls')),
    )

... then some of the views that returned a redirect to a URL
constructed with string formatting would redirect to a path
that didn't include the 'mapit/' prefix.  If these paths are
constructed with reverse() instead, it works properly.

This also introduces some new URL pattern names, so the URL
patterns can be referred to more concisely.


dracos:
:+1:

PR #229, opened 2016-04-01
Allow overriding of email subject prefix

sihugh:
Offering this as it's a change we've made to our fork.

In GOV.UK we want to specify different email subject lines for MapIt emails in each environment.  This change allows the EMAIL_SUBJECT_PREFIX to be set in general.yml.

The space after the closing ']' in the example is intentional in order to format the full email subject nicely when it includes following text.


dracos:
Thanks :)

PR #236, opened 2016-05-06
Fix reversing of ONS coded polygon URLs

dracos:


mhl:
Looks good to me :+1:

The only comment I have is incidental (i.e. not actually to do with the lines that have been changed) but I think it would be easier to read the `area_code_lookup` function if `area_code` were renamed to `area_code_type` and then the `area_id` parameter were renamed to `area_code`.

mhl:
(and sorry about missing this in my previous PR...)

PR #241, opened 2016-05-25
[UK] May 2016 Boundary-Line import.

### LINKED ISSUE #240: Import May 2016 Boundary-Line

dracos:
Deal with the Welsh Assembly boundary issues caused by new data being
provided too early, a returning parish council that presumably should
not have been deleted, and a duplicate unit ID.

Fixes #240

### DIFF
--- a/mapit_gb/controls/2016-05.py
+++ b/mapit_gb/controls/2016-05.py
@@ -1,0 +1,81 @@
+# A control file for importing May 2016 Boundary-Line. This control file should
+# be okay to run as a first import file also, it uses previous Boundary-Line
+# information if present to fix up issues with the source data.
+
+from mapit.models import Area, Generation
+
+
+def code_version():
+    return 'gss'
+
+
+def check(name, type, country, geometry, ons_code, commit, **args):
+    """Should return True if this area is NEW, False if we should match against ONS code,
+       or an Area to be used as an override instead"""
+
+    if type in ('WAC', 'WAE'):
+        # We might be in the situation where we have bad data, as OS provided
+        # incorrect boundaries from October 2013 to May 2015 inclusive:
+        # https://web.archive.org/web/20150922064012/https://www.ordnancesurvey.co.uk/business-and-government/help-and-support/products/boundary-line.html
+        # The October 2015 MapIt import put the old boundaries in again as new
+        # creations. Let's move the old boundaries to be new, as that's what
+        # they are now, and push back the current boundary generation to cover
+        # where they were.
+        try:
+            name = name.replace(' Assembly Const', '').replace(' Assembly ER', '')
+            current = Area.objects.get(type__code=type, name=name, generation_high=Generation.objects.current())
+            old_but_new = Area.objects.get(
+                codes__type__code='gss', codes__code=ons_code, generation_high__lt=Generation.objects.current())
+            assert current.generation_low.id == old_but_new.generation_high.id + 1
+            print "* Wales change, before: %d %s (%d-%d), %d %s (%d-%d)" % (
+                old_but_new.id, old_but_new.name, old_but_new.generation_low_id, old_but_new.generation_high_id,
+                current.id, current.name, current.generation_low_id, current.generation_high_id)
+            current.generation_low = old_but_new.generation_low
+            if commit:
+                current.save()
+            old_but_new.generation_low = Generation.objects.new()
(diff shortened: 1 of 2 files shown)


mapit_gb/controls/2016-05.py
```diff
@@ -0,0 +1,73 @@
...
+    # have ONS codes and so can be ignored/ detected that way:
+    #
+    # Bristol           2015/1871
+    # Cherwell          2015/1872
+    # Colchester        2015/1859
+    # Elmbridge         2016/301
+    # Exeter            2016/65
+    # Gloucester        2015/2026
+    # Knowsley          2015/2036
+    # Lincoln           2015/1461
```
mhl: Going by the Boundary-Line appendix, shouldn't this list include:

```
CRAVEN 2012/3150 
CHERWELL 2016/1872 
WEST OXFORDSHIRE 2012/2993
```

mapit_gb/controls/2016-05.py
```diff
@@ -0,0 +1,73 @@
...
+    """Should return True if this area is NEW, False if we should match against ONS code,
+       or an Area to be used as an override instead"""
+
+    if type in ('WAC', 'WAE'):
+        # We might be in the situation where we have bad data, as OS provided
+        # incorrect boundaries from October 2013 to May 2015 inclusive.
+        # October 2015 put the old boundaries in again as new creations.
+        # Let's move the old boundaries to be new, as that's what they are now,
+        # and push back the current boundary generation to cover where they were.
+        try:
```
mhl: This method looks good to me given the situation you've explained in the comment - have the OS documented that mistake somewhere?

mapit_gb/management/commands/mapit_UK_import_boundary_line.py
```diff
@@ -59,14 +59,11 @@ def handle_label(self, filename, **options):
             ons_code = feat['CODE'].value if feat['CODE'].value not in ('999999', '999999999') else None
             unit_id = str(feat['UNIT_ID'].value)
             area_code = feat['AREA_CODE'].value
-            if hasattr(control, 'patch_boundary_line'):
-                patch = control.patch_boundary_line(name, ons_code, area_code)
```
mhl: I'm probably being dense here, but going through the history, did any control file modules ever have a `patch_boundary_line` function? (It clearly looks safe to remove, of course, I just wondered if it was ever used.)

mapit_gb/controls/2016-05.py
dracos: Cherwell is already there (it's 2015/1872, not 2016, another mistake in OS output :) ); Craven and W. Oxfordshire aren't Electoral Change Orders (rewarding the council) but only related alterations due to parish boundary changes, which are generally very minor. Sometimes they get new ONS codes, sometimes they don't, it doesn't really matter to us either way (this time it looks like they do). Can list them for completeness, sure.

mhl:
This looks good to me, but I haven't tried the import on a copy of the live database yet. I'm trying that now...

mapit_gb/management/commands/mapit_UK_import_boundary_line.py
dracos: 2015-10 temporarily did before it was realised that didn't work if you wanted your first import to be 2015-10 as you couldn't use that control file as it relied on previous imports. So no, nothing uses it and I imagine in future it will be the same, any patches will have to be here in case it's the first import (or the control files are written to work in either scenario, as actually I have done for 2016-05!).

mapit_gb/controls/2016-05.py
dracos: They did on the product support page at release, but have removed it now there's a new release. archive.org has it at https://web.archive.org/web/20150922064012/https://www.ordnancesurvey.co.uk/business-and-government/help-and-support/products/boundary-line.html

mhl: Ah, OK - it'd be nice to include that link in the comment, I think.

mhl:
The import seemed to work fine - after `Area.objects.filter(type__code='16W').delete()`, here's the output of the dry run:

https://gist.github.com/mhl/49846fadedc840fb57efd3ca7c37cf20

... and then with `--commit`:

https://gist.github.com/mhl/1818a3db0ce6dcc80c47dc946e7630d5

So :+1: as far as I can tell :)

PR #243, opened 2016-06-15
[UK] Improved postcode validation.

dracos:
This is much quicker, and includes a missing "P" in the third character list. Thanks to Andy Gimblett for suggesting these changes to the https://github.com/mysociety/uk-postcode-utils package.

### DIFF
--- a/mapit_gb/countries.py
+++ b/mapit_gb/countries.py
@@ -2,6 +2,7 @@ import re
 
 from django.core.urlresolvers import reverse, NoReverseMatch
 from django.http import HttpResponseRedirect
+from ukpostcodeutils import validation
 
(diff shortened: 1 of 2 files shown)


dracos:
It seemed like you might well not want to treat those as valid, or have a different list, so left the library to accept whatever "extra" you decided. It could be a list stored there, I guess, but didn't seem worth it.

stevenday:
> It seemed like you might well not want to treat those as valid, or have a different list, so left the library to accept whatever "extra" you decided. It could be a list stored there, I guess, but didn't seem worth it.

I dunno, it's exactly the sort of thing I'd be delighted to find out a postcode library had built in. Obviously it might need some options or something - I'll make a ticket on postcodeutils.

PR #244, opened 2016-06-28
Update 'add_ons_to_gss' command to support CSVReader for Python 2

davidbasalla:
The command would error when running with Python 2.7, failing to recognise the Welsh characters `ô` and `â` with "UnicodeEncodeError: 'ascii' codec can't encode character u'\xf4' in position 16: ordinal not in range(128)". This is due to `csv.reader` changing between Python 2 and 3 - in 2 it expected a file-like object on which next() would return raw bytes, and in 3 one which would return unicode strings (found by @jennyd). We opt to explicitly check which version of Python is running and run the old code for < Python 3.

We also stop assigning `name` for the 'code-change.csv' as it is not used anywhere else in the script and the encoding probably won't be handled correctly with Python 2. We stop assigning `type` and `name` for 'missing-codes.csv' for the same reasons.

### DIFF
--- a/mapit_gb/management/commands/mapit_UK_add_ons_to_gss.py
+++ b/mapit_gb/management/commands/mapit_UK_add_ons_to_gss.py
@@ -4,6 +4,7 @@
 import codecs
 import csv
 import os.path
+import sys
 from django.core.management.base import NoArgsCommand
 from mapit.models import Area, CodeType
 from psycopg2 import IntegrityError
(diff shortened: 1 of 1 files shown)


dracos:
Hi, thanks very much for this :) It's not a script we turn any more upstream, so hadn't spotted this.

I have pushed a refactor fixup commit at 2e341a0c2b09a7d709288254b42f71cec0b6ceba which I think simplifies this somewhat – could you check that it is still okay for you? If so, please do squash the two commits together in your branch. Could you also wrap your commit message to the standard 50/72 line lengths; thank you!

dracos:
This has been merged as d659d40 (with my change), thanks!

PR #248, opened 2016-08-15
Django 1.10 support

dracos:
Move management commands to argparse, improve some very old code that could be done much more easily, improve test output.


stevenday:
👍 Looks good to me

PR #255, opened 2016-08-19
Support jessie/xenial in install scripts.

dracos:
The script hardcoded a check for trusty as a proxy for PostGIS 2.
Instead, look up the postgis package version installed and use that.

### DIFF
--- a/bin/install-as-user
+++ b/bin/install-as-user
@@ -68,11 +68,16 @@ fi
 echo -n "Setting up database... "
 if ! psql -l | egrep "^ *$DB_NAME *\|" > /dev/null
 then
-    if [ trusty = "$(lsb_release -s -c)" ]
+    VERSION_POSTGIS=$(dpkg-query -W -f '${Version}\n' postgresql-*-postgis*|sort -rV|head -1)
+    if [ 2.0 = "$(echo -e '2.0\n'$VERSION_POSTGIS | sort -V | head -1)" ]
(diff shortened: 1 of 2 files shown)


bin/install-as-user
```diff
@@ -68,11 +68,13 @@ fi
 echo -n "Setting up database... "
 if ! psql -l | egrep "^ *$DB_NAME *\|" > /dev/null
 then
-    if [ trusty = "$(lsb_release -s -c)" ]
+    VERSION_POSTGIS=$(dpkg-query -W -f '${Version}\n' postgresql-*-postgis*|sort -rV|head -1)
+    POSTGIS_TWO=$(echo -e '2.0\n'$VERSION_POSTGIS | sort -V | head -1)
```
mhl: This is neat :) However I'd probably either:

* Add a comment to explain that `POSTGIS_TWO` will be '2.0' if the PostGIS version is >= 2.0, and the actual version otherwise.
* Or set `POSTGIS_TWO` to `Yes` or `No`, e.g. with:

```shell
if [ 2.0 = "$(echo -e '2.0\n'$VERSION_POSTGIS | sort -V | head -1)" ]
then POSTGIS_TWO=Yes
else POSTGIS_TWO=No
fi
```

mhl:
A small optional comment, otherwise looks good :+1:

dracos:
Discovered this was running somewhere where `-e` didn't work, switched to `printf`.

PR #260, opened 2016-08-22
Only try and set gdal exceptions on first use.

dracos:
If it is done too early, at module import, it triggers a bug in Django whereby the DataSource drivers aren't registered and an import fails.

(I wondered if it was simpler to force a DataSource registration in e.g. manage.py, but as this is the only place using osgeo.gdal, it seemed to make sense to put it there, and nothing front-end uses DataSource anyway. Happy to do it differently if others prefer.)


mhl:
:+1:

PR #273, opened 2016-11-15
Fix display of Geometry object in admin.

### LINKED ISSUE #272: Encoding error on geometry admin page

dracos:
Fixes #272.

### DIFF
--- a/mapit/models.py
+++ b/mapit/models.py
@@ -5,7 +5,7 @@ from django.contrib.gis.db import models
 from django.conf import settings
 from django.db import connection
 from django.db.models.query import RawQuerySet
-from django.utils.encoding import python_2_unicode_compatible
+from django.utils.encoding import python_2_unicode_compatible, smart_text
 
 from mapit import countries
 from mapit.geometryserialiser import GeometrySerialiser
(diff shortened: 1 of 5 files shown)


mapit/models.py
```diff
@@ -333,7 +333,7 @@ class Meta:
         verbose_name_plural = 'geometries'
 
     def __str__(self):
-        return '%s, polygon %d' % (self.area, self.id)
+        return '%s, polygon %d' % (smart_text(self.area), self.id)
```
dracos: Q: Why didn’t you use `u''` as an easier way to fix this?
A: Django 1.8, still supported, still supports Python 3.2, therefore so do we. `u''` was only re-added in 3.3.

stevenday:
Looks good, but looking at a the models.py this is in, there are a few other models that take some text either from a data source or the user and return it from `__str__`, should we smart_text those too?

dracos:
Any text coming from a CharField will be a Unicode string, so they’re all already covered (I think I did check them all, double check helpful :) ); the reason this issue arose is because the Geometry `__str__` used `self.area` directly, which caused the area’s `__str__` function to be called and return a byte string in Python 2.

stevenday:
aha, gotcha. I double-checked and I think you're right :shipit:

PR #274, opened 2016-11-15
Fix use of old Django aggregate functions

dracos:
Tweak/add couple of tests to show that these functions were still being called, and stop using them.


stevenday:
👍

Issue #127, opened 2014-06-09
Allow grouping of areas as one concept


dracos:
What we really want is for the Area model to be called Geometry (the Geometry model exists because old versions of postgres were very bad at multipolygon PinP; it could be called Polygon I guess), and a new Area model to exist which can refer to multiple Geometrys. This would allow e.g. use of OpenStreetMap relation IDs to consistently refer to an area, and allow council boundary changes in the UK (as opposed to ward boundary changes within a council) to maintain the same identifier for the council more easily.

dracos:
One alternative/possibility is that we could already have this grouping, in that areas can have parents, with no generational constraints otherwise. So you could have:

* Area: Glasgow City Council, generations 1-30, *no associated geometry*, has e.g. local-authority-eng code.
    * Area: Glasgow City Council, generations 1-14 (mapit.mysociety.org/area/141651), has GSS code
        * Areas of the wards of Glasgow
    * Area: Glasgow City Council, generations 15-30 (mapit.mysociety.org/area/2579), has GSS code
        * Areas of the wards of Glasgow

`/point` or `/postcode` return almost the same as now, because the 'parent' area has no associated geometry – you'll have an extra (not present) parent ID in the council area that you can use/ignore. This would rely on users of the data knowing that this existed for local authorities and storing that identifier if they wished to, but that's no worse than present!

One big (the main?) user that has this issue is FixMyStreet – it would certainly be possible to move to these parent identifiers if they existed, and this would provide greater stability for the next council GSS ID change (when because we *didn't* want to change FixMyStreet, we bodged MapIt to put the new boundary in the old ID and the old boundary in a new ID, I know).

If this solution was acceptable, it would be much more straightforward than anything else.

dracos:
Except what about someone wanting the wards of Glasgow, holding Glasgow's 'stable' ID? They should be able to call children on it and get them, surely. So that leads to:
* Area: Glasgow City Council, generations 1-30, *no associated geometry*, has e.g. local-authority-eng code.
    * All areas of the wards of Glasgow, whatever generation (the children call lets you look up previous generations with a parameter)
    * Area: Glasgow City Council, generations 1-14, has GSS code
    * Area: Glasgow City Council, generations 15-30, has GSS code

Which means that someone making a children call for Glasgow (with no type filter) would get two Glasgow, which makes little sense to a casual user, or is it okay? It's probably not okay.

torotil:
[...]

PR #275, opened 2016-12-05
Local authority eng codes

symroe:
This PR adds codes from the [local-authority-eng](https://local-authority-eng.register.gov.uk/) register.

I checked the map file in to the data directory as the remote location of that file isn't stable, even though the content is.

The thing I'm least sure about is 06cc1ff. Guessing at the CodeType from a regex seems to work for the other IDs, but a three letter code isn't unique to `local-authority-eng` (for example, we have `local-authority-wls` coming, with the same ID system).

One option would be to convert the query to an `__in` query, with each regex block appending to the list of CodeTypes. This feels like a fairly large change with possible performance implications, so I thought it better to add it as it's own commit that I can remove if it's too tricky to get done in this PR.


dracos:
Thanks! There is a better lookup to use if you have a code, e.g. `/code/local-authority-eng/AMB` – the fact `/area/` lets you use ONS/GSS codes on top of ID is historical, and we probably don't want to add any new things to that.

symroe:
Oh cool! I use `/area/[gss]` _all_ the time, never knew about `/code/[CodeType]/[Code]`.

I'll remove 06cc1ff and force push this branch.

mapit/models.py
```diff
@@ -386,7 +386,7 @@ class CodeType(models.Model):
     # This could be extended to a more generic data store of information on an
     # object, perhaps.
 
-    code = models.CharField(max_length=10, unique=True, help_text="A unique code, eg 'ons' or 'unit_id'")
+    code = models.CharField(max_length=100, unique=True, help_text="A unique code, eg 'ons' or 'unit_id'")
```
dracos: As type.code and code.code both went up to 500 when they were migrated, shall we just do the same here for consistency? And we might as well do nametype.code up to 500 as well, then they all are :)

mapit_gb/data/gss-to-local-authority-eng.tsv
```diff
@@ -0,0 +1,445 @@
...
+E06000013	local-authority-eng:NLN	North Lincolnshire
+E07000147	local-authority-eng:NNO	North Norfolk
+E06000024	local-authority-eng:NSM	North Somerset
+E08000022	local-authority-eng:NTY	North Tyneside
+E07000218	local-authority-eng:NWA	North Warwickshire
+E07000134	local-authority-eng:NWL	North West Leicestershire
+E10000023	local-authority-eng:NYK	North Yorkshire
+E07000154	local-authority-eng:NOR	Northampton
+E10000021	local-authority-eng:NTH	Northamptonshire
+E06000057	local-authority-eng:NBL	Northumberland
```
dracos: Should this file also have other GSS codes that match to the same local authority? e.g. here, E06000048 is also Northumberland.

dracos: There are 8 councils with multiple GSS codes - Glasgow, East Dunbartonshire, St Albans, Welwyn Hatfield, East Hertfordshire, Stevenage, Northumberland, Gateshead.

dracos: This of course then raises another issue as two areas in MapIt would have the same code, and `/code/` currently returns an error when that happens. No reason it couldn't return a list if multiple results, though...

dracos: #127 would presuambly 'solve' that issue, if it allowed the concept to have its own IDs (of which this would certainly be one), separate from its boundaries (which have the GSS IDs).

mapit_gb/management/commands/mapit_UK_import_local-authority-eng-codes.py
```diff
@@ -0,0 +1,43 @@
...
+        code_type, _ = CodeType.objects.update_or_create(
+            code='local-authority-eng',
+            defaults={
+                'description': "GOV.UK local-authority-eng codes"
+            }
+        )
+
+        # Add Codes
+        data_path = os.path.abspath(
+            os.path.join('mapit_gb', 'data', 'gss-to-local-authority-eng.tsv'))
```
dracos: This means it only works if you run manage.py from inside its own directory? Could it use `__file__` in some way so that it doesn't matter - there are some other uses in this repo.

mapit_gb/management/commands/mapit_UK_import_local-authority-eng-codes.py
```diff
@@ -0,0 +1,43 @@
...
+            if register_code_type == 'local-authority-eng':
+                try:
+                    gss_code = Code.objects.get(code=line['gss'])
+                    Code.objects.get_or_create(
+                        code=code,
+                        type=code_type,
+                        area=gss_code.area
+                    )
+                except Code.DoesNotExist:
+                    pass
```
dracos: Presumably we want to know if we're missing a GSS code? :)

symroe: How do you want to know? Is there some logging system, or is printing to stdout ok?

mapit_gb/data/gss-to-local-authority-eng.tsv
symroe: Ah, sorry, this is because in my install I didn't have `E06000048` (I've not run one of the fix up commands)

symroe: Hmm, I can't seem to run those commands anyway, for other reasons (`mapit_UK_fix_2011-10` and `mapit_UK_fix_2012-05` have import errors, `mapit_UK_fix_2013-10` needs an input file, but I don't know what it is, etc). Anyway, I'll accept that there is a problem here, but I'm not sure if the fix should be in this PR? Doesn't every other code type suffer from this too?

dracos: No, I imagine those commands only work at the precise point they were written for, I'm afraid (or some other reason). Yes, other code types would, but we don't currently have any non-unique codes in the system, as far as I know. (Actually, I guess OLF/OLG share ONS IDs, but they can be looked up via historical `/area/`.)

mapit_gb/management/commands/mapit_UK_import_local-authority-eng-codes.py
symroe: Ok, I've made that change :)

mapit_gb/data/gss-to-local-authority-eng.tsv
dracos: Ah, right, yes, so the example I gave was the same code in areas of two different types (OLF and OLG). Probably easiest for this for now if we just put them on the "latest" GSS codes (ie. what you have already done) and then raise a new ticket to do something about it? This will have a deadline of as soon as ONS issue a new GSS code for a local authority...

symroe: That's quite soon! But I agree, this should be a different issue. Are you happy to write it, as you will have a better idea of how to express it. I'm happy to have a go at working on it if you like.

dracos: Is it? They don't do it very often, or we'd probably have had to do something properly sooner! :) I've added a comment on #127 whereby we could add parent areas to the councils that remain stable through such boundary changes, then these codes could theoretically be placed on there instead...

symroe: Oh sorry, I'm thinking about wards, where there are 13 counties getting new ward codes next May. That doesn't change the council geography GSS codes though, so we're fine.

PR #280, opened 2017-02-09
UI refresh

zarino:
This brings the generic UI improvements from https://github.com/mysociety/mapit.mysociety.org/issues/29 into mapit core.

Works really well with the latest version of https://github.com/mysociety/mapit.mysociety.org/pull/32, but needs testing in the other mapit cobrands (like global, and italy).


mapit_it/templates/mapit/base.html
```diff
@@ -1,88 +1,4 @@
-{% load static from staticfiles %}
+{% extends "mapit/base.html" %}
```
dracos: Drive-by: This won't work I'm afraid because it's the same template being overridden.
Put the description in a separate template and then override that?

zarino: The only reason I'm overriding `base.html` here is to insert the translated page description. Could we just upload the translation via Transifex instead, and then the override wouldn't even be required?

dracos: It already is `trans`, so yes,i don't think this is necessary at all.

mapit/static/mapit/sass/modules/_footer.scss
```diff
@@ -1,74 +0,0 @@
...
-
-  hr {
-    border-width: 1px 0 0 0;
-  }
-
-  p {
-    margin-top: 0;
-  }
-
-  .mapit-copyright {
```
dracos: Could we keep the copyright text being a bit smaller? :)

mapit/static/mapit/sass/modules/area_view.scss
```diff
@@ -1,94 +0,0 @@
...
-        clear: none;
-        float: right;
-        padding: 1em 0.625em 2.5em 0.625em;
-    }
-*/
-
-    @media (min-width: 60em) {
-        float: right;
-        clear: none;
-        width: 28.75em;
```
dracos: I quite liked the two columns of text output on an area page.

mapit/static/mapit/sass/_forms.scss
```diff
@@ -0,0 +1,103 @@
...
+    width: auto;
+    margin: 0 0.5em;
+    vertical-align: middle;
+}
+
+.homepage-try {
+    input[type="text"] {
+        @extend .form-control;
+        @extend .form-control--inline;
+        text-transform: uppercase;
```
dracos: On Global, "LATITUDE,LONGITUDE" does not fit.

mapit/static/mapit/sass/modules/area_list.scss
```diff
@@ -1,46 +0,0 @@
...
-            margin: 0;
-            font-size: 0.75em;
-        }
-    }
-}
-
-// Creates area_level_1 to area_level_10
-// optionally add to list items to add indentation suggesting hierarchy
-
-@while $area_level > 0 {
```
dracos: It would be a shame to lose this (which indents the area lists on e.g. http://global.mapit.mysociety.org/point/4326/8.55,47.366667.html).

dracos:
One other thing, do you think the line between sections of the docs might be a bit faint? I can only just see it on my monitor :)

mapit/static/mapit/sass/modules/_footer.scss
zarino: Done!

mapit/static/mapit/sass/modules/area_view.scss
zarino: Done!

mapit/static/mapit/sass/_forms.scss
zarino: Done!

mapit/static/mapit/sass/modules/area_list.scss
zarino: Done!

PR #284, opened 2017-02-15
Block for extra content in area/postcode info

zarino:
Useful for cobrands that want to display a small amount of custom data, or a message to users, without overriding the whole template.

(We’re using this in MapIt UK to display a message about business users and paid plans.)

### DIFF
--- a/mapit/templates/mapit/postcode.html
+++ b/mapit/templates/mapit/postcode.html
@@ -8,21 +8,22 @@
 <header class="area_info">
     <h2>{{ postcode.postcode }}</h2>
     <ul>
-{% if postcode.coordsyst %}
-<li>{% ifequal postcode.coordsyst "G" %}OSGB{% else %}IRE65{% endifequal %}
-E/N: {{ postcode.easting }}, {{ postcode.northing }}
-{% endif %}
-{% if postcode.wgs84_lat or postcode.wgs84_lon %}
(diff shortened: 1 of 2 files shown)


mapit/templates/mapit/postcode.html
```diff
@@ -8,21 +8,22 @@
...
-{% endif %}
+        {% endblocktrans %}</li>
+      {% endif %}
+        {% url json_view postcode=postcode.postcode as json_url %}
+        <li>{% blocktrans %}Get <a href="{{ json_url }}">this data as JSON</a>{% endblocktrans %}</li>
     </ul>
-    {% url json_view postcode=postcode.postcode as json_url %}
-    <p>{% blocktrans %}Get <a href="{{ json_url }}">this data as JSON</a>{% endblocktrans %}</p>
+
+    {% block area_info_extra %}{% endblock %}
```
dracos: Should this be `postcode_info_extra`?

zarino: I considered it, but then the HTML element it appears in is called `area_info` (even on the postcode page) – so I stuck with `area_info_extra` because it's extra content to go in the `.area_info` element.

dracos: Yeah, I noticed that after (and merged+deployed as you see) :)

PR #286, opened 2017-02-22
Fixing reverse for form_submitted()

### LINKED ISSUE #285: Redirect error from front page postcode form
(Version: current master, UK configuration and data, brand new install.)

Something (and I am unable to tell exactly what, sorry) causes the front-and-centre postcode lookup form POSTed to mapit.views.postcodes.postcode to fail, and redirect incorrectly to a 404 for

http://mapithost.domain.tld/postcode/mapit.views.postcodes.postcode

Possibly a missing reverse() function on the urlconf?

handelaar:
Using django.http.HTTP.HttpResponseRedirect() and django.urls.reverse()

Would fix #285 (does for me anyway)

### DIFF
--- a/mapit/views/postcodes.py
+++ b/mapit/views/postcodes.py
@@ -148,7 +148,7 @@ def form_submitted(request):
     pc = request.POST.get('pc', None)
     if not request.method == 'POST' or not pc:
         return redirect('/')
-    return HttpResponseRedirect(reverse('mapit-postcode', kwargs={ 'postcode':pc, 'format':'html'}))
+    return HttpResponseRedirect(reverse('mapit-postcode', kwargs={'postcode': pc, 'format': 'html'}))
 
 
 @ratelimit(minutes=3, requests=100)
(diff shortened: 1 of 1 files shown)


handelaar:
Fixed the travis error for Flake8. Remaining problems with checks are.... a bit old.

dracos:
All `django.shortcuts.redirect` does is call `reverse`, as you can see at https://github.com/django/django/blob/master/django/shortcuts.py#L150
So I would assume that doing nothing but changing 'mapit.views.postcodes.postcode' to 'mapit-postcode' would also fix the issue, and not have any older-Django-version issues :) I have now found in the Django 1.10 release notes: "The ability to reverse() URLs using a dotted Python path is removed." so that's the issue.

It would be good to have a test added showing that this fails in 1.10/11 – Travis automatically runs the tests on 1.8-1.11. Django 1.8 is supported until early 2018, and so we support it also.

PR #287, opened 2017-02-23
Fix redirect notation for Django 1.10+

handelaar:
As per discussion under #286.


handelaar:
Fixes #285

dracos:
Thanks! I've added a test for the change, and also stripped non-alphanumerics and fixed one other test. I'll merge assuming Travis is okay :)

PR #294, opened 2017-03-27
Add OSM ref attribute as part of global import

chrismytton:
## What does this do?

This adds the `ref` attribute from OSM, if present, as a new kind of `Code` when performing a global import from KML files.

## Why was this needed?

This is part of the work adding Mapit as a source of area data in EveryPolitician - https://github.com/everypolitician/everypolitician/issues/595. This addition should make matching between Mapit/OSM and Wikidata simpler, which in turn will make it easier to reconcile the data between the two in EveryPolitician.

For a full explanation, see @mhl's comment - https://github.com/everypolitician/everypolitician/issues/595#issuecomment-288382893.

### DIFF
--- a/mapit_global/fixtures/global.json
+++ b/mapit_global/fixtures/global.json
@@ -24,5 +24,6 @@
     { "pk": 20, "model": "mapit.type", "fields": { "code": "OWA", "description": "OSM Political Boundary (Ward)" } },
 
     { "pk": 1, "model": "mapit.codetype", "fields": { "code": "osm_rel", "description": "OSM relation" } },
-    { "pk": 2, "model": "mapit.codetype", "fields": { "code": "osm_way", "description": "OSM way" } }
+    { "pk": 2, "model": "mapit.codetype", "fields": { "code": "osm_way", "description": "OSM way" } },
+    { "pk": 3, "model": "mapit.codetype", "fields": { "code": "osm_attr_ref", "description": "OSM reference attribute" } }
 ]
--- a/mapit_global/management/commands/mapit_global_import.py
+++ b/mapit_global/management/commands/mapit_global_import.py
@@ -303,6 +303,17 @@ class Command(LabelCommand):
                     if old_lang_codes:
                         verbose('Removing deleted languages codes: ' + ' '.join(old_lang_codes))
                     m.names.filter(type__code__in=old_lang_codes).delete()
+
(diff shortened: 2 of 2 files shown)


chrismytton:
@dracos Thanks for the pointers! I've changed the code to be more like the name handling bits, so that now deals with the `ref` changing or being removed.

chrismytton:
@dracos Thanks! I haven't deployed MapIt before, can I be cheeky and ask you to deploy this for me? 🙊 🤞

dracos:
Sure (ftr it's `sudo mysociety vhost global.mapit.mysociety.org` on milo/stork, then I manually loaded the fixture with `python manage.py loaddata global`, doesn't look like the deploy does that); is the plan to do an entire new import, or have a one-off script to add the refs to the existing areas?

dracos:
(The current KML used for the last import is in `/home/matthew/global-2017-02` on milo.)

chrismytton:
> is the plan to do an entire new import, or have a one-off script to add the refs to the existing areas?

Ooh, erm, not sure! Is it easy to do an entire import? Or would it be better (FSVO) if I wrote a script to do the refs individually?

dracos:
"Easy" in technical terms, the process is https://wiki.mysociety.org/wiki/Global_MapIt#Updating_boundaries – run script on firefly, run script on copy to test, run script on live – but it will take days to run in total (even if you skipped the copy step). The data will be a couple of months newer. I don't know if we have an opinion on how often/much data we want to import, we probably should have done something different about dealing with out-of-date boundaries. A script to add the refs to the current should be straightforward, and will certainly get you results quicker.

mhl:
Wouldn't it be possible for @chrismytton to do what I suggested on everypolitician/everypolitician#595, i.e.

> Then (@dracos may correct me on this) you could create a new inactive generation in MapIt, run mapit_global_import just on the directory with OCL boundaries, run mapit_generation_raise_on_current_areas and then activate the new generation.

dracos:
Yes, that would work (assuming the existing area comparison works a bit better than the UK postcode import which says most things have changed when they haven't...)! Actually, you could tweak the script to allow it to also run on an existing live generation (I did that manually for adding `OCL`) and then not need a new generation that was only adding new metadata.

mhl:
I made a PR with that change here: https://github.com/mysociety/global.mapit.mysociety.org/pull/2

PR #295, opened 2017-04-04
Move MapIt global code to its own repo.

dracos:
This code should all now be present in https://github.com/mysociety/global.mapit.mysociety.org

There is one issue outstanding, the area sorting function in `add_codes`. The easiest thing to do would be to reinstate the two lines of `mapit/countries/__init__.py` that this PR removes and then it would work as long as the user had `mapit_global/countries.py` existing (which it does). Is there a nicer way?


mhl:
I think there's a `mapit_global` reference to be removed from the `MANIFEST.in` as well.

mhl:
In the new repository's `general.yml-example`, shouldn't the `COUNTRY` be `Global`, and the `AREA_SRID` `4326`?

PR #299, opened 2017-05-08
[UK] Use database for co-ordinate transformation.

dracos:
If the application is running on a server with GDAL 1.9 (and maybe 1.8),
its OSGB co-ordinate transformation can be up to a few metres out. As we
use the transformation by the database when saving new coordinates, also
use it when fetching so we can match and display accurate information.

### DIFF
--- a/mapit/management/commands/mapit_import_postal_codes.py
+++ b/mapit/management/commands/mapit_import_postal_codes.py
@@ -131,17 +131,12 @@ class Command(LabelCommand):
             pc = Postcode.objects.get(postcode=self.code)
             if location:
                 if pc.location:
-                    curr_location = (pc.location[0], pc.location[1])
+                    curr_location = pc.location
+                    # Postcode locations are stored as WGS84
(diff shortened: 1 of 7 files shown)


mhl:
Looks good to me - just a couple of things:

* Do you have a link to a bug report describing the problem where GDAL 1.9 is returning wrong results? It would be helpful to include this in the commit message if so. (e.g. I think GDS are using GDAL 1.10, and that might help them to assess if this affect them.)
* I'd probably add a test which adds a `Postcode` for somewhere in Great Britain, runs it through `as_uk_grid` and checks the eastings / northings output is as expected. (And the same for a postcode in Northern Ireland.)

dracos:
I've added a link to the GDAL trac ticket and tests (the GB one I've validated against other data, the NI one doesn't agree but as GDAL + postgres match I'm going to go with that anyway). I've then fixed some other test thing (order mattered and tests failed on firefly) and a co-ord display issue.

PR #301, opened 2017-07-13
Add packages file for Debian Stretch

mhalden:
This PR adds a packages file with the updated packages for Debian Stretch.


dracos:
Thanks for this!

Now that I look at this, I'm wondering if we can actually simplify things by only installing `postgis` – modern distribution versions (since jessie/xenial) will then install the needed versions (previously they were only a suggests or not mentioned at all). I've added a new commit to this PR that does this and introduces a `packages.generic` file, could you see if that works okay for you?

dracos:
Haven't heard back, but I've squashed this together, as it should be fine.

PR #302, opened 2017-10-03
Add support for a 'country' filter in query parameters

### LINKED ISSUE #55: allow look-ups by name that are restricted by area (e.g. a particular country)
At the moment you can search for areas by a prefix (e.g. with `/areas/cambr.html`) but one can't restrict this search to a particular country. This might be useful to some people.

mhl:
This change means that you can add `?country=E` (say) to queries for
multiple areas to restrict the results to those areas that have a
Country with code `E`.

This should be particularly helpful for global.mapit.mysociety.org
if we set an appropriate Country for each Area, since one is almost
always interested in results from a particular country.

Fixes #55

### DIFF
--- a/mapit/views/areas.py
+++ b/mapit/views/areas.py
@@ -12,7 +12,7 @@ from django.conf import settings
 from django.shortcuts import redirect, render
 from django.views.decorators.csrf import csrf_exempt
 
-from mapit.models import Area, Generation, Geometry, Code, Name
+from mapit.models import Area, Generation, Code, Name
 from mapit.shortcuts import output_json, output_html, output_polygon, get_object_or_404, set_timeout
 from mapit.middleware import ViewException
 from mapit.ratelimitcache import ratelimit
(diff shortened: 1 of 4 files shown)


mhl:
n.b. The test failures here are due to changes in the Travis build environment - I've made a pull request with fixes for those errors [here](https://github.com/mysociety/mapit/pull/303) so once that's merge, this should be rebased over that.

mapit/views/areas.py
```diff
@@ -358,29 +363,13 @@ def areas_by_point(request, srid, x, y, bb=False, format='json'):
...
+        args['polygons__polygon__bbcontains'] = location
     else:
-        if method == 'box':
-            args['polygons__polygon__bbcontains'] = location
-        else:
-            geoms = list(Geometry.objects.filter(polygon__contains=location).defer('polygon'))
-            args['polygons__in'] = geoms
-        areas = Area.objects.filter(**args)
+        geoms = list(Geometry.objects.filter(polygon__contains=location).defer('polygon'))
+        args['polygons__in'] = geoms
```
dracos: To continue this, these two lines could now be `args['polygons__polygon__contains'] = location` I assume?

PR #304, opened 2017-10-12
[UK] October 2017 Boundary-Line import.

### LINKED ISSUE #219: Scilly isles boundaries
Our wards don't have boundaries, http://mapit.mysociety.org/area/14334.html but they all exist as parishes which do: http://mapit.mysociety.org/area/31021.html - the scilly script could presumably copy the boundaries across rather than use TR postcodes.

dracos:
Hopefully comments in control file explain things (need expanding if not!).
Fixes https://github.com/mysociety/mapit.mysociety.org/issues/64
Fixes #219


mapit_gb/controls/2017-10.py
```diff
@@ -0,0 +1,59 @@
...
+
+from mapit.models import Area, Generation
+
+
+def code_version():
+    return 'gss'
+
+
+def check(name, type, country, geometry, ons_code, commit, **args):
+    """Should return True if this area is NEW, False if we should match"""
```
davea: Could this comment be expanded to describe why it would return an Area instead of True/False?

mapit_gb/controls/2017-10.py
```diff
@@ -0,0 +1,59 @@
...
+
+    current = Generation.objects.current()
+
+    # May 2017 Boundary-Line gave Clydebank Central the wrong ID,
+    # return the relevant row if present
+    if type == 'UTW' and ons_code == 'S13003126':
+        return Area.objects.get(codes__type__code='gss', codes__code='S13003136')
```
davea: Should this be wrapped in a `try`/`except` like the ones below, or will it definitely always exist?

mapit_gb/controls/2017-10.py
```diff
@@ -0,0 +1,59 @@
...
+            if commit:
+                a.codes.update_or_create(type__code='gss', defaults={'code': 'W05000186'})
+            return a
+        except Area.DoesNotExist:
+            pass
+
+    # These areas were incorrectly named in May 2017 Boundary-Line,
+    # so find the misnamed entries if present
+    if type == 'CED' and name == 'Uckfield North ED':
+        return Area.objects.get(
```
davea: Ditto the `try`/`except` here, in case (for whatever reason) someone's skipped May 2017?

mapit_gb/controls/2017-10.py
dracos: If the area doesn't exist, it'll raise an exception to create a new area, which is what we want and what would happen anyway (as the same thing would happen when it tried to look up S13003126).

mapit_gb/controls/2017-10.py
dracos: Both of these were new in May 2017 (hence them getting their names wrong), so if May 2017 isn't present, same as above, a not exists exception will be raised, as it would have been anyway a bit later on.

mapit_gb/controls/2017-10.py
dracos: Ah yes, that comment is a bit out of date, the one in 2016-05.py is better!

PR #305, opened 2017-10-24
First stab at fixing /nearest

davea:
Something about the recent move to Postgres 9.6 and PostGIS 2.3 caused
the /nearest query to take much longer than 10 seconds to complete.

This commit updates the query to use the ‘<->’ centroid distance operator
instead, which makes better use of the geoindex and is much faster to
execute.

Handily the work @dracos did to introduce the TrigramDistance function
into Django 1.10 was easily adapted for this purpose, as it uses the
same operator.

This needs a bit of tidying and refactoring before merging really.


dracos:
Tweaked after merge to transform to 4326.

PR #309, opened 2017-11-22
 Add .map.html endpoint to show map of areas

### LINKED ISSUE #307: feature request: add some easy way of visualizing areas returned by `/areas` queries
This would be useful for the Democratic Commons project's audit of boundary data, which includes trying to see how much boundary data that that we'd need is already complete / accurate enough in OpenStreetMap. The idea is to make it easy to get a quick at-a-glance sense of what the coverage at a particular admin level is for given countries in global.mapit . For example, we might compare this visualization with other maps of the administrative boundaries for a country. (In order to actually do this, we also need to merge and deploy #302, and wait for the job which is currently running to set the country on all areas to complete.)

This could be a simple Leaflet-based page that takes a MapIt JSON results URL from `/areas`, fetches a GeoJSON file for each of areas in those results, and displays it on the map. However, I think this is useful enough that perhaps it's worth building into MapIt - my idea was that as well as the `.html` extension for the text-based results for multiple areas, you could also have `.map.html`, which would render the areas returned on a map.

I guess we would probably want to limit the number of areas that would be fetched, since, for example, queries for all areas whose name starts with `a` would produce a browser-crashing number of areas to render.

dracos:
This works for any list page (children/point/areas by name/etc). It restricts itself to the first 200 areas of the list if there are more.
As list pages are output streamed, it has to collect the IDs as it goes and do the map at the end; I think this should be okay for any sensible page.
Fixes #307

### DIFF
--- a/mapit/shortcuts.py
+++ b/mapit/shortcuts.py
@@ -1,4 +1,5 @@
 import itertools
+import re
 
 from django import http
 from django.db import connection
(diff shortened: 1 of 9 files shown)


chrismytton:
Here's a screenshot of this in action:

screen shot 2017-11-23 at 09 21 08

dracos:
"Would be good to add this feature to the documentation" - yeah, I thought we'd leave it until it'd been used in anger a bit and see if there are any unforeseen issues. It could even become part of the standard /areas/ HTML page for a smaller number of areas, perhaps.

PR #310, opened 2017-11-22
Add many-many table for additional countries.

### LINKED ISSUE #308: change the Area <-> Country relationship to be many-to-many
This might seem counter-intuitive, but there are some important cases where an area can be in more than one country, but it's still useful to be able to query on all the areas associated with a particular country. The primary use case we have for this is trying to add country data to areas in global.mapit - there are [many areas that overlap with multiple admin_level 2 boundaries](https://gist.github.com/mhl/9ebc7c454ab7a5c7da15eceb23a1db00), particularly where a region is contested. For example, Taiwan / China, Russia / Ukraine, etc. At the moment we're setting these areas to a special country with code `?` to indicate multiple enclosing countries, but this means they won't be returned by queries for any of the enclosing countries.

This is a politically sensitive thing to model, of course, and it would be best if we could stick to the principle that we just represent what's in OSM as accurately as possible - in these cases I think that means showing multiple countries.

(Another example of where this might be useful would be if accurate postcode polygons are ever released as open data for the UK - there are lots of postcodes that straddle country boundaries within the United Kingdom.)

One concern that @dracos pointed out when we discussed this in person was that we would have to be careful not to break the existing API - this means making sure that the `country` and `country_name` properties in the returned JSON remain the same in cases where there's a single country (and perhaps falling back on the `?` country code in new ambiguous cases. Then perhaps we could a new `countries` array like:

```
"countries": [
  {
     "code": "CN",
     "name": "China"
   },
   {
     "code": "TW",
     "name": "Taiwan"
   }
]
```

or separate `countries` and `countries_names`, or whatever...

dracos:
These countries are included in JSON output in a `countries` list.
Fixes #308 - in a different way, adds m2m table on top of foreignkey rather than replacing it, as I thought that was easier to avoid a large migration and the cases where single country is all that is needed. Importer can decide which either/both to use how they like.


dracos:
"The HTML version of area pages doesn't display the .countries values at all, but it would be nice if it did." Oh yes! I've added a `list_countries` method that combines the two entries and shows them all (I guess we might want to keep a global override to not show "Global", but meh).

dracos:
The query is right, I think, because you do want it to be in either. The LEFT OUTER joins are also fine (one from area to country, one from area to the m2m join table, one from the m2m join table to country). The issue is that for an entry matching on area.country that also has multiple 'countries' entries, it's returning one row per countries entry, but we only need one of those results. I've added some `distinct()`s which I think fixes what you've raised, and I can't think of any issues with that (might have missed something!).

mhl:
I think my only concern with adding the `distinct()` would be that it potentially could make the query a lot slower, but I guess we'd have to measure that difference.

PR #313, opened 2017-12-05
Some map.html and other improvements

dracos:
This loads areas as they come in rather than once at the end, which enables hover/popup info.
Also upgrade Leaflet, stop caching templates in development and improve rate limiter configuration.


project/settings.py
```diff
@@ -200,13 +196,6 @@
             'mapit.context_processors.country',
             'mapit.context_processors.analytics',
         ),
-        # List of callables that know how to import templates from various sources.
-        'loaders': (
-            ('django.template.loaders.cached.Loader', (
-                'django.template.loaders.filesystem.Loader',
-                'django.template.loaders.app_directories.Loader',
-            )),
-        ),
```
mhl: I'm sure I'm missing something simple, but how is it that this stops caching of templates in development only, and doesn't remve the cached loader entirefly?

dracos: No, don't think you're missing anything, whoops!

dracos: Ah, no, I wasn't! From Django 1.11, the cached loader is automatically enabled when DEBUG is False.

mhl: Ah, OK - thanks :)

dracos:
Yeah, we'll need to remove the existing script (added in 45f7ab4e) too. I've pushed a fixup to stop it highlighting on area page, thanks.

PR #319, opened 2018-03-13
Manually prefetch any matching many-many countries

dracos:
iterator() and prefetch() can't be used together, so the prefetch of many-many
countries is not happening in area lists, causing a slowdown if many areas are
to be returned. Instead, as we already do with codes, we manually fetch all of
the relevant countries from the join table and store them on the model objects
as we iterate through them.

### DIFF
--- a/mapit/models.py
+++ b/mapit/models.py
@@ -6,6 +6,7 @@ from django.conf import settings
 from django.db import connection
 from django.db.models.query import RawQuerySet
 from django.utils.encoding import python_2_unicode_compatible, smart_text
+from django.utils.functional import cached_property
 
 from mapit import countries
 from mapit.geometryserialiser import GeometrySerialiser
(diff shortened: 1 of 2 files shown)


mapit/models.py
```diff
@@ -265,17 +265,30 @@ def as_dict(self, all_names=None):
...
         return out
 
     def list_countries(self):
-        countries = [c.name for c in self.countries.all()]
+        countries = [c['name'] for c in self.all_m2m_countries]
         if self.country:
             countries.append(self.country.name)
         return countries
 
+    @property
```
davea: Could this use [`django.utils.functional.cached_property`](https://docs.djangoproject.com/en/2.0/ref/utils/#django.utils.functional.cached_property) instead? It'd make the code simpler and remove the need for the setter as the decorator handles that.

dracos:
Edited post merge to include blank results too, as we want a negative cache result as well (the only result on UK, after all).

PR #328, opened 2018-08-14
Remove bytestrings from old migrations

tomkins:
Currently under Python 3 - Django wants to create new migration files to fix the change in description from a byte string to unicode string.

This PR updates models.py to ensure all strings are unicode strings, and updates the first migration to remove byte strings.


dracos:
Thanks very much this and apologies for the delay in getting it merged!

PR #333, opened 2019-01-25
Docker deployment

KhadijaMahanga:
This PR enables developers to develop, deploy and run mapit on docker containers. 

Files added:
- `Dockerfile`
- `docker-compose.yml`
- `docker-entrypoint.sh`
- `runtume.txt`
- `CHECKS`

Files modified:
- `project/settings.py`
- `project/ urls.py`  - to serve static files


sagepe:
Thanks for your interest in MapIt and for submitting a pull request!

We are adding Docker support to some of our projects, so it was interesting to see how you have approached this.

However, I'm sorry to say that we can't accept this PR in its current format.

We'd prefer a slightly different approach to the structure of the `Dockerfile` with the aim of reducing layers and final image size. For example `RUN` line can be combined when installing dependencies, recommended packages should be omitted unless necessary and apt metadata and cache can be removed following installation.

The PR contains files that are not directly related to the Docker build or the compose file but rather concerned with particular deployment environments (eg the `CHECKS` file). It would be better if these were explained in dedicated commits or omitted entirely.

The commits also vary in content in that some extend functionality and some contain fix-ups to previous commits, often without any real explanation. At the PR stage, we'd prefer that fix-ups were rebased into the original commits.

If you are interesting in additional feedback, please let me know.

DavidLemayian:
Thanks @sagepe .

An in-line review would be great and we'll make the changes ASAP.

Also, we can definitely squash/rebase the commits in the branch to clean up commits. Are there any guidelines on commit messages?

And lastly, I notice vagrant is still the choice of the day. Would it be more useful to just open an issue and let the core team prioritise the switch to Docker so as not to mess up the code / documentation?

PR #337, opened 2019-04-21
Fix allow_tags

DavidLemayian:
This has been removed from 2.0 - https://docs.djangoproject.com/en/2.2/internals/deprecation/#deprecation-removed-in-2-0


dracos:
Thanks very much for spotting this! I have added some more information to the commit message to explain the change.

PR #340, opened 2019-06-20
Use unary_union instead of cascaded_union.

torotil:
cascaded_union was deprecated in Django 1.10 and has been removed in 2.0.

The current Django requirement is >= 1.11.2 which is still fine.


dracos:
Thanks for this! I have also got the doctests in that file to be run by the test suite, whoops, which would have caught this.

PR #353, opened 2020-03-06
[UK] Script for 2020 England council changes.

dracos:
This script is to be run at the start of April to create the new Buckinghamshire Council and its new wards.

### DIFF
--- a/mapit_gb/management/commands/mapit_UK_add_2020_structural_changes.py
+++ b/mapit_gb/management/commands/mapit_UK_add_2020_structural_changes.py
@@ -1,0 +1,73 @@
+# coding=utf8
+
+from __future__ import print_function
+from django.core.management.base import BaseCommand
+from mapit.models import Area, Generation, Type, Country, NameType, CodeType
+
+
+class Command(BaseCommand):
+    help = "Sort out boundaries for 2020 council changes"
+
+    def add_arguments(self, parser):
+        parser.add_argument('--commit', action='store_true')
+
+    def handle(self, *args, **options):
+        self.commit = options['commit']
+        self.g = Generation.objects.current()
+        self.gn = Generation.objects.new()
+        self.name_type = NameType.objects.get(code='O')
+        self.code_type = CodeType.objects.get(code='gss')
+        self.country = Country.objects.get(code='E')
+
+        self.raise_generation_on_everything_else()
+
+        self.areas = Area.objects.filter(generation_low_id__lte=self.g, generation_high_id__gte=self.g)
+
+        self.get_existing()
+        self.create_buckinghamshire()
+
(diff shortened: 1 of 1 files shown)


mapit_gb/management/commands/mapit_UK_add_2020_structural_changes.py
```diff
@@ -0,0 +1,72 @@
...
+        self.name_type = NameType.objects.get(code='O')
+        self.code_type = CodeType.objects.get(code='gss')
+        self.country = Country.objects.get(code='E')
+
+        self.raise_generation_on_everything_else()
+
+        self.areas = Area.objects.filter(generation_low_id__lte=self.g, generation_high_id__gte=self.g)
+
+        self.get_existing()
+        self.buckinghamshire()
```
davea: I think this line will crash because `self.buckinghamshire` isn't a method any more - `get_existing` has set that instance variable to an `Area`. I guess renaming the method to `create_buckinghamshire` or similar is required.

mapit_gb/management/commands/mapit_UK_add_2020_structural_changes.py
```diff
@@ -0,0 +1,72 @@
...
+        m = Area(
+            name=name, type=Type.objects.get(code=typ), country=self.country,
+            generation_low=self.gn, generation_high=self.gn,
+        )
+        if self.commit:
+            m.save()
+            m.names.update_or_create(type=self.name_type, defaults={'name': name})
+            if gss:
+                m.codes.update_or_create(type=self.code_type, defaults={'code': gss})
+            m.polygons.create(polygon=geom)
```
davea: Running locally with `--commit`, this line crashes with
```
psycopg2.errors.CheckViolation: new row for relation "mapit_geometry" violates check constraint "enforce_srid_polygon"
DETAIL:  Failing row contains (996813, 163793, 0103000020E6100000010000009E3C0000EDF23EFC23E7ECBF15565944DFC549...).
```

dracos: I don't understand this one, the polygon has just come out of the db so how can it be a different SRID? :-/ What is your DB srid set to?

dracos: Running on server (without commit), the geom's srid appears to be 27700 when fetched so should be okay on saving. I guess if your db mismatched your general.yml this could happen? But if it's consistent, will have to keep looking.

davea: Sigh, I had it `AREA_SRID: 4326` in general.yml. Sorry for the wild goose chase. Works fine now!

Issue #365, opened 2020-11-02
October 2020 Boundary-Line 


dracos:
- [x] Put in our cache
- [x] Import into mapit.mysociety.org

dracos:
First pass at October 2020 Boundary-Line - Buckinghamshire Council's wards are in as CEDs, not UTWs, and all the Buckinghamshire district wards are still present (the district councils are not). I have told OS about this here: https://twitter.com/dracos/status/1323303337090027521
Only other thing is the GLA now appears to have a GSS code at last, so need to match that up with existing GLA entry if we want (or just let it import as a new, though identical, area).

dracos:
So we always need to ignore all the Bucks district wards, they shouldn't be in there. And try and match the GLA area to the existing one if it is present. For the wards of Bucks Council itself, we need to treat them as UTWs, not CEDs - but they still won't match up because there are no GSS codes, so need to go on their name somehow, ugh.

dracos:
OS have confirmed this is deliberate because of the Situation they are considering the old wards/EDs to still exist (as elections have not yet taken place for them), although the district councils have gone and the new unitary council does exist. Will have to think how to deal with this from our own internal point of view.

dracos:
If you are importing this dataset into an empty database, I think it should be fine, apart from find_parents will presumably die because it cannot find the parents of the Buckinghamshire CEDs and DIWs. We could add a special case to find_parents so that it maps those CEDs and DIWs to the Buckinghamshire UTA. This should then work fine, I *think* (even the shortcut output, it will have county/district both with the same parent area which is correct).

If you are importing this dataset into an existing database with previous Boundary-Line in, and you haven't run the script from April to update Buckinghamshire, then I think it should also be okay (CEDs don't have IDs and can't recall how that works, but worst case is it creates some duplicate entries) with the same issue as above.

[...]

PR #367, opened 2020-11-03
October 2020 Boundary-Line import

dracos:
See #365.
Adds a control file that matches up the GLA area and allows CED/DIW through as new areas if inside a Buckinghamshire UTA.
And allows `find_parents` to accept a UTA parent for a CED/DIW also because of Buckinghamshire.


davea (approved):
👍

PR #368, opened 2020-11-03
Django 3 support

dracos:
Minimum python 3 is 3.6 for 3.0.

We need to stop using vendored six, as that has been removed, and the
URL dispatcher has changed so it no longer passes in non-matching parts
(in our case, the format ending).

Hopefully that is all.


davea (approved):
3️⃣ ✅  That looks nice and simple :)

PR #373, opened 2021-01-29
Improve caching of postcode endpoints

theseanything:
This PR modifies the responses for the postcode endpoint to not use a streaming HTTP response, to allow Django to cache the responses preventing the need for subsequent expensive database queries.

Previously, the output_json method always returned a streaming HTTP response, which prevented the need to load the full response contents into memory before sending, lowering memory usage. However, this prevented Django from caching views using the output_json method. This modifies the output_method to not use a streaming response for those responses already loaded into memory (i.e. not using the iterdict) and would gain no benefit. This will allow for Django to automatically cache those views and reduce the need for expensive database queries (e.g. mainly the /postcodes endpoints).

### DIFF
--- a/mapit/shortcuts.py
+++ b/mapit/shortcuts.py
@@ -66,10 +66,10 @@ def output_json(out, code=200):
         content = map(lambda x: x, content)
 
         response = http.StreamingHttpResponse(content_type='application/json', status=code)
-        response['Cache-Control'] = 'max-age=2419200'  # 4 weeks
         attr = 'streaming_content' if getattr(response, 'streaming', None) else 'content'
         setattr(response, attr, content)
 
+    response['Cache-Control'] = 'max-age=2419200'  # 4 weeks
     response['Access-Control-Allow-Origin'] = '*'
 
     return response
--- a/mapit/shortcuts.py
+++ b/mapit/shortcuts.py
@@ -54,20 +54,18 @@ def output_json(out, code=200):
 
     if type(out) is dict:
         response = http.JsonResponse(
-            out, status=code, encoder=GEOS_JSONEncoder, json_dumps_params=json_dumps_params
-        )
+            out,
+            status=code,
+            encoder=GEOS_JSONEncoder,
+            json_dumps_params=json_dumps_params)
     else:
         encoder = GEOS_JSONEncoder(**json_dumps_params)
         content = encoder.iterencode(out)
 
-        # We don't want a generator function (iterencode) to be passed to an
-        # HttpResponse, as it won't cache due to its close() function adding it to
-        # an instance attribute.
-        content = map(lambda x: x, content)
-
-        response = http.StreamingHttpResponse(content_type='application/json', status=code)
-        attr = 'streaming_content' if getattr(response, 'streaming', None) else 'content'
-        setattr(response, attr, content)
+        response = http.StreamingHttpResponse(
+            streaming_content=content,
+            content_type='application/json',
+            status=code)
 
     response['Cache-Control'] = 'max-age=2419200'  # 4 weeks
     response['Access-Control-Allow-Origin'] = '*'
--- a/mapit/shortcuts.py
+++ b/mapit/shortcuts.py
@@ -49,26 +49,29 @@ def output_json(out, code=200):
         if isinstance(out, dict):
             out['debug_db_queries'] = connection.queries
         indent = 4
-    encoder = GEOS_JSONEncoder(ensure_ascii=False, indent=indent)
-    content = encoder.iterencode(out)
 
-    # We don't want a generator function (iterencode) to be passed to an
-    # HttpResponse, as it won't cache due to its close() function adding it to
-    # an instance attribute.
-    content = map(lambda x: x, content)
+    json_dumps_params = {'ensure_ascii': False, 'indent': indent}
 
-    types = {
-        400: http.HttpResponseBadRequest,
-        404: http.HttpResponseNotFound,
-        500: http.HttpResponseServerError,
-    }
-    response_type = types.get(code, http.StreamingHttpResponse)
+    if type(out) is dict:
+        response = http.JsonResponse(
+            out, status=code, encoder=GEOS_JSONEncoder, json_dumps_params=json_dumps_params
+        )
+    else:
+        encoder = GEOS_JSONEncoder(**json_dumps_params)
+        content = encoder.iterencode(out)
 
-    response = response_type(content_type='application/json; charset=utf-8')
+        # We don't want a generator function (iterencode) to be passed to an
+        # HttpResponse, as it won't cache due to its close() function adding it to
+        # an instance attribute.
+        content = map(lambda x: x, content)
+
+        response = http.StreamingHttpResponse(content_type='application/json', status=code)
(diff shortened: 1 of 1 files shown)


mapit/shortcuts.py
```diff
@@ -50,25 +50,24 @@ def output_json(out, code=200):
...
         indent = 4
     encoder = GEOS_JSONEncoder(ensure_ascii=False, indent=indent)
-    content = encoder.iterencode(out)
 
-    # We don't want a generator function (iterencode) to be passed to an
-    # HttpResponse, as it won't cache due to its close() function adding it to
-    # an instance attribute.
-    content = map(lambda x: x, content)
+    if type(out) is dict:
+        response = http.JsonResponse(out, encoder=GEOS_JSONEncoder, status=code)
```
dracos, line 55: This loses us the indent on staging, and changes the behaviour on non-ASCII characters - the latter shouldn't matter for anything reading it, of course, and the former can be got around easily enough, but it would be good to remain consistent between the two types of output. I think you could pass in a `json_dumps_params` argument.
`response = http.JsonResponse(out, encoder=GEOS_JSONEncoder, status=code, json_dumps_params=json_dumps_params)`

mapit/shortcuts.py
```diff
@@ -50,25 +50,24 @@ def output_json(out, code=200):
             out['debug_db_queries'] = connection.queries
         indent = 4
     encoder = GEOS_JSONEncoder(ensure_ascii=False, indent=indent)
```
dracos, line 52: This encoder is now only used in the new else so can be moved to inside it.

dracos: Perhaps could have a `json_dumps_params = {'ensure_ascii': False, 'indent': indent}` here, then can use that in both paths.

mapit/shortcuts.py
```diff
@@ -50,25 +50,24 @@ def output_json(out, code=200):
...
-        404: http.HttpResponseNotFound,
-        500: http.HttpResponseServerError,
-    }
-    response_type = types.get(code, http.StreamingHttpResponse)
+        # We don't want a generator function (iterencode) to be passed to an
+        # HttpResponse, as it won't cache due to its close() function adding it to
+        # an instance attribute.
+        content = map(lambda x: x, content)
+
+        response = http.StreamingHttpResponse(content_type='application/json; charset=utf-8', status=code)
```
dracos, line 64: For consistency with Django's `JSONResponse`, do we want to drop the (superfluous) charset here to match?

dracos (changes requested):
Thanks for looking at this! Makes sense we don't use a StreamingHttpResponse when it's not necessary. I guess we've not had to worry about this as we have varnish sat in front of Django and it seems to be caching all the streamed responses okay (though looks like it is buffering them, which it shouldn't be, hmm, one for another day). I've made a couple of small suggestions just for consistency between the two paths, hope that's okay, otherwise looks good :)

theseanything:
Hi @dracos, thanks for you comments and your review! Just for context, the motivation for this PR is that we use the API to power other user facing applications, and wanted to use the existing Django caching mechanism rather than setting up another caching layer between applications. I noticed that we had very low hit rate in memcache, as the bulk of our requests are the to postcode json endpoint. So this PR will mean we actually start utilising memcached, with added advantage of in future being able to use a shared memcached to further improve caching across multiple machines.

I'm not sure about was the cache control header, wonder if we should make the caching times consistent between streaming and non-streaming responses? I think for all responses the max-age is being set as the [CACHE_MIDDLEWARE_SECONDS](https://github.com/mysociety/mapit/blob/e55ae067b118b87e207b303d1c118b600f0fba3c/project/settings.py#L61) ([Django docs](https://docs.djangoproject.com/en/3.1/topics/cache/#the-per-site-cache)). But overridden for streaming responses explicitly `response['Cache-Control'] = 'max-age=2419200'`. Should make CACHE_MIDDLEWARE_SECONDS = 2419200, and remove setting the header separately?

dracos:
Yes, I think it would be good for the cache time to be consistent too, so I would move that line down to be with the Access-Control-Allow-Origin. If you remove setting the header separately, it will not use CACHE_MIDDLEWARE_SECONDS instead because the cache middleware does not fire on streaming responses.

We could change this hard-coded figure to read the CACHE_MIDDLEWARE_SECONDS variable (or a separate variable so they can remain different) and use that, I imagine, though I'd probably want that as a separate PR to this.

With the current live code, the caching is:
* HTML for front page, postcode results page, most pages - Cache-Control/memcached for CACHE_MIDDLEWARE_SECONDS
* HTML for area list pages - no caching header, no memcached
* Most JSON/geometry output - Cache-Control for 4 weeks (hardcoded in MapIt code), no memcached.
* JSON output for area lists - Cache-Control for 4 weeks (hardcoded in MapIt code), no memcached.

With the PR as it is when I write this, that would become:
* HTML - same
* Most JSON/geometry output - Cache-Control/memcached for CACHE_MIDDLEWARE_SECONDS
* JSON output for area lists - Cache-Control for 4 weeks (hardcoded in MapIt code), no memcached.

If you move that line down as suggested, then the caching behaviour becomes:
* HTML - same
* Most JSON/geometry output - Cache-Control/memcached for 4 weeks (hardcoded in MapIt code)
* JSON output for area lists - Cache-Control for 4 weeks (hardcoded in MapIt code), no memcached.

PS I realise I don't think we actually link to http://mysociety.github.io/coding-standards.html#code-reviews anywhere here, but please note we'd prefer fixup commits because then it is easier to review the changes you've made based on feedback, rather than having to work it out within the already checked code, thanks :)

theseanything:
Thanks @dracos! I've moved the setting the header to be consistent for both non-streaming and streaming responses. So now the behaviour is as you described:

> - HTML - same
> - Most JSON/geometry output - Cache-Control/memcached for 4 weeks (hardcoded in MapIt code)
> - JSON output for area lists - Cache-Control for 4 weeks (hardcoded in MapIt code), no memcached.


 Sorry I was confused before and you're right the streaming responses don't use the CACHE_MIDDLEWARE_SECONDS const.

Happy to look at a future PR to make the cache control max-age configurable.

dracos:
Thanks! Looks good. Now looking at it overall, I think the following diff can be applied:

```diff
-        # We don't want a generator function (iterencode) to be passed to an
-        # HttpResponse, as it won't cache due to its close() function adding it to
-        # an instance attribute.
-        content = map(lambda x: x, content)
-
-        response = http.StreamingHttpResponse(content_type='application/json', status=code)
-        attr = 'streaming_content' if getattr(response, 'streaming', None) else 'content'
-        setattr(response, attr, content)
+        response = http.StreamingHttpResponse(
+            streaming_content=content,
+            content_type='application/json',
+            status=code)
```

We no longer need the do-nothing map, because we're always using a StreamingHttpResponse at this point, and we no longer need to check whether it is streaming or not, because it always is. This simplifies it quite a bit. If you want to commit this change to the branch, go ahead and I can merge the PR tomorrow, or I can apply it tomorrow at the same time, just letting you know in case you spot I'm doing anything wrong!

dracos (approved):
Thanks for your work on this and feeding it upstream :)

PR #375, opened 2021-03-15
Add within parameter to point lookup.

### LINKED ISSUE #363: Add lookup of areas near a point
Our `/point` endpoint (areas covering a point) could also be described as "all the areas within 0m of a point", so extending that with a parameter to look up areas within Nm of a point could be useful. Or it could be added to the `/nearest` endpoint.

dracos:
Fixes #363

### DIFF
--- a/mapit/templates/mapit/api/point.html
+++ b/mapit/templates/mapit/api/point.html
@@ -63,6 +63,13 @@
                                 <i>country</i>, to restrict results to areas with particular country codes (multiple country codes separated by commas).
                             {% endblocktrans %}
                         </li>
+                      {% if within_maximum %}
+                        <li>
+                            {% blocktrans trimmed %}
+                                <i>within</i>, to show results within the specified distance (up to {{ within_maximum }} {{ area_srid_units }}) of the point, not just the point itself (not /box).
+                            {% endblocktrans %}
+                        </li>
+                      {% endif %}
                     </ul>
                 </dd>
 
--- a/mapit/views/areas.py
+++ b/mapit/views/areas.py
@@ -394,8 +394,18 @@ def areas_by_point(request, srid, x, y, bb=False, format=''):
 
     q = query_args(request, format)
 
+    max_within = getattr(settings, 'MAPIT_WITHIN_MAXIMUM', 0)
+    try:
+        within = float(request.GET.get('within', 0))
+        if within < 0 or within > max_within:
+            raise ValueError
+    except ValueError:
+        raise ViewException(format, _('Bad "within" parameter specified'), 400)
+
     if method == 'box':
         q &= Q(polygons__polygon__bbcontains=location)
+    elif within:
+        q &= Q(polygons__polygon__dwithin=(location, within))
     else:
         q &= Q(polygons__polygon__contains=location)
     areas = Area.objects.filter(q).distinct()
--- a/conf/general.yml-example
+++ b/conf/general.yml-example
@@ -18,6 +18,10 @@ COUNTRY: 'GB'
 # Optional, defaults to 4326.
 AREA_SRID: 27700
 
+# Set this to the maximum distance (in AREA_SRID units) allowed for the within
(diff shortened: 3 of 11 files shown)


mapit/views/areas.py
```diff
@@ -394,8 +394,15 @@ def areas_by_point(request, srid, x, y, bb=False, format=''):
 
     q = query_args(request, format)
 
+    try:
+        within = float(request.GET.get('within', 0))
+    except ValueError:
+        raise ViewException(format, _('Bad within specified'), 400)
```
davea, line 400: This error is a bit terse and hard to parse, something like `'Bad "within" parameter specified'` would be clearer.

mapit/views/areas.py
```diff
@@ -394,8 +394,15 @@ def areas_by_point(request, srid, x, y, bb=False, format=''):
 
     q = query_args(request, format)
 
+    try:
+        within = float(request.GET.get('within', 0))
```
davea, line 398: We should validate that this is >= zero, as negative values appear to cause an `InternalError: Tolerance cannot be less than zero` SQL error. Might be an idea to clamp it to an upper bound too, I'm not sure if a really high figure could otherwise be used to get every single area and if that would be a problem?

davea (changes requested):
A couple of little things to make it easier and more robust is all.

mapit/views/areas.py
dracos: I'll make it a config variable so we can adjust as necessary.

mapit/templates/mapit/api/point.html
```diff
@@ -63,6 +63,11 @@ <h3>{% blocktrans trimmed %}
                                 <i>country</i>, to restrict results to areas with particular country codes (multiple country codes separated by commas).
                             {% endblocktrans %}
                         </li>
+                        <li>
+                            {% blocktrans trimmed %}
+                                <i>within</i>, to show results within the specified distance of the point, not just the point itself (not <kbd>/box</kbd> version).
```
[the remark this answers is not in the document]
dracos: Used a context processor and assumed 4326 or not, which the rest of the code also does.

mapit/views/areas.py
```diff
@@ -394,10 +394,13 @@ def areas_by_point(request, srid, x, y, bb=False, format=''):
 
     q = query_args(request, format)
 
+    max_within = getattr(settings, 'MAPIT_WITHIN_MAXIMUM', 0)
     try:
         within = float(request.GET.get('within', 0))
+        if within < 0 or within > max_within:
```
davea, line 400: This will unfortunately break the  feature entirely if `within` has a value > 0 and `MAPIT_WITHIN_MAXIMUM` is undefined (because `max_within` defaults to zero).

mapit/views/areas.py
```diff
@@ -394,10 +394,13 @@ def areas_by_point(request, srid, x, y, bb=False, format=''):
 
     q = query_args(request, format)
 
+    max_within = getattr(settings, 'MAPIT_WITHIN_MAXIMUM', 0)
```
davea, line 397: Would be good to document this in `general.yml-example`.

davea (changes requested):
A small bit of finessing required but almost there.

mapit/views/areas.py
dracos: Yes, I don't think it should work unless you specify a maximum (for the reason you stated, that it would be possible to overload by requesting all areas with a huge distance). So it's a way of switching the feature off too.

mapit/views/areas.py
dracos: I was going to add/document it in mapit.mysociety.org's conf/general.yml and pull it into its settings there, iyswim, but yes, might as well be consistent here too. Shouldn't really have any conf/general.yml stuff in this project.

davea (approved):
Woo!

PR #376, opened 2021-03-18
Include JSON information in GeoJSON properties.

dracos:


davea (approved):
A test wouldn't hurt, but it's small enough that it probably doesn't matter :)

PR #384, opened 2022-05-04
[UK] May 2022 Boundary-Line import.

dracos:
A couple of parishes back from the dead, and an incorrect ward code.

### DIFF
--- a/mapit_gb/controls/2022-05.py
+++ b/mapit_gb/controls/2022-05.py
@@ -1,0 +1,29 @@
+# A control file for importing May 2022 Boundary-Line.
+
+# This control file assumes previous Boundary-Lines have been imported,
+# because it uses that information. If this is a first import, use the
+# first-gss control file.
+
+from mapit.models import Area, Generation
+
+
+def code_version():
+    return 'gss'
+
+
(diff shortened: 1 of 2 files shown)


mapit_gb/controls/2022-05.py
```diff
@@ -0,0 +1,29 @@
...
+    an ONS code, or an Area to be used as an override instead."""
+
+    missing_parishes = {
+        'Wexham Court CP': 'E04001211',
+        'Britwell CP': 'E04012334',
+    }
+    if type == 'CPC' and name in missing_parishes.keys():
+        # Missing in previous editions
+        a = Area.objects.get(codes__type__code='gss', codes__code=missing_parishes[name])
+        a.generation_high = Generation.objects.current()
```
davea, line 25: By jumping `generation_high` from 35 to 46 does this mean the area would be included in the intervening generations? Does that matter?

dracos: Yes, it would be, and no I don't think it matters because I assume these parishes didn't cease to exist for the past 3 years and they've just been missing incorrectly ever since then.

davea (approved):

PR #388, opened 2022-05-17
Django 3.2/4 testing, dropping older things

dracos:

PR #389, opened 2022-05-18
[UK] Add command to add GSS codes to CED areas using CSV from ONS

davea:
Current data from ONS: https://geoportal.statistics.gov.uk/documents/ons::county-electoral-division-to-county-may-2021-lookup-for-england/about (needs converting to CSV manually, didn't really want to introduce a dependency for parsing XLSX files)


dracos (approved):
Looks good 🗺️

dracos:
Note the 2.7 flake8 test broke as the script uses a python 3 only bit of code. I know we're dropping py2 in another PR (soon presumably) so perhaps we could just drop the py2 flake8 in Actions, there's still a py3 run of it.

davea:
Oops, yes. The `{}` set literal syntax and/or f-strings I guess. Should I drop py2 on `master` now or leave it until #388 lands?

dracos:
Might as well drop it now so we don't have the failing test, even if only for a (hopefully) short time :)
Looks like I didn't drop six in #388 anyway so more could be done there.

PR #392, opened 2022-06-29
Do not log our own ViewExceptions.

dracos:
Since Django 2.1, all 500s are logged as errors, including ones manually
generated by the code. We do not want to receive logged errors for ones
we are choosing to generate, so mark the response as already logged with
the logger's internal flag.


sagepe (approved):
Looks good to me, seems that setting this will get picked up in log_response in django.utils.log.

PR #395, opened 2022-11-01
Do not log our own JSON ViewExceptions.

dracos:
Following on from b5172f6d, make sure JSON error output is also marked as already logged, as we do not want to receive logged errors for ones we are choosing to generate.


davea (approved):
I'm assuming there's no need for this to be applied to the `StreamingHttpResponse` on the other side of the `else`, is there?

dracos:
No, that's only ever used for successful outputs, any error call to output_json (with a code provided) provides an error dict.

PR #399, opened 2023-03-27
Add filtering and excluding by mutliple types and countries in raise generation script.

neprune:
Allows selecting multiple countries / types e.g. `--type "ABC" --type "DEF" --country "E" -- country "W"`.

Also allows flipping to exclude by type or country via `--type-mode "all-but" --country-mode "all-but"`.


dracos (approved):
Thanks, should come in handy in May :)

dracos:
(PS Tiny typo in commit message before merging)

Issue #401, opened 2023-04-09
HTML OSM Administrative Boundary Levels jumbled, JSON OK


jidanni:
The HTML levels are always out of order (e.g., 8, 4, 6, 2).
The JSON levels are OK.
```
u=https://global.mapit.mysociety.org/point/4326/-87.782,42.166
{ w3m -dump $u.html;
  w3m -dump_source $u|json_pp;}|grep Level

    ID 999545, OSM Administrative Boundary Level 8 (O08)
    ID 959228, OSM Administrative Boundary Level 4 (O04)
    ID 974962, OSM Administrative Boundary Level 6 (O06)
    ID 958719, OSM Administrative Boundary Level 2 (O02)
      "type_name" : "OSM Administrative Boundary Level 2"
      "type_name" : "OSM Administrative Boundary Level 4"
      "type_name" : "OSM Administrative Boundary Level 6"
      "type_name" : "OSM Administrative Boundary Level 8"
```

dracos:
The HTML and JSON are both output in alphabetical order by name - the `json_pp` program sorts keys on output by default. I agree the sorting isn't great for the HTML output, this is a duplicate for https://github.com/mysociety/global.mapit.mysociety.org/issues/5 - there's a little fix there we could do in this repo.

PR #405, opened 2023-05-04
Boundary Line May 2023

dracos:
Skip the ward that doesn't exist, and skip all county council electoral divisions because Suffolk's 2025 changes are present.

Also fixes https://github.com/mysociety/mapit/issues/401 (a fix for https://github.com/mysociety/global.mapit.mysociety.org/issues/5).


davea (approved):
👍

PR #407, opened 2023-05-25
Faster ONSPD import

dracos:
This switches from the default loop/check/save to postgres copy into a temporary table and update/insert based on differences. The first run compared to previous way will have a lot of minute updates due to the location calculation being slightly different (going from CSV E/N -> 4326 in DB rather than code), but in the order of millimetres(!). Import time goes from 12 hours to a few minutes.

### DIFF
--- a/mapit/iterables.py
+++ b/mapit/iterables.py
@@ -1,3 +1,6 @@
+import io
+
+
 def defaultiter(it, default):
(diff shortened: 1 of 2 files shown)


davea (approved):
⚡ ⏩ 🎉  This is super fast now!

PR #413, opened 2023-11-06
Fix MapIt install script

mikejamesthompson:
This PR attempts to fix the automated installation script for MapIt on more recent versions of Ubuntu (but not Jammy, 22.04) and Debian Linux.

I have confirmed the installation succeeds on clean installs of Ubuntu Focal (20.04) and Debian Bullseye (11) on Amazon EC2 instances at the default hostname, and tested subsequent loading of boundary data works on Debian Bullseye.

It **does not appear to work automatically on Ubuntu Jammy (22.04)** because of changes in `apt-get install` behaviour, I think related to the [`needrestart`](https://askubuntu.com/questions/1367139/apt-get-upgrade-auto-restart-services) command, but I haven't had time to investigate this properly.

To use this version of the install script, you can use the override behaviour in the `install-site.sh` script. Run the following commands:

```
curl -O https://raw.githubusercontent.com/mysociety/commonlib/master/bin/install-site.sh
chmod u+x install-site.sh
export REPOSITORY_URL_OVERRIDE=https://github.com/mikejamesthompson/mapit
export BRANCH_OVERRIDE=fix-install-script
sudo --preserve-env ./install-site.sh --default mapit mapit
```

Changes are explained in commit messages.

### DIFF
--- a/bin/site-specific-install.sh
+++ b/bin/site-specific-install.sh
@@ -23,7 +23,7 @@ misuse() {
 [ -z "$DISTRIBUTION" ] && misuse DISTRIBUTION
 [ -z "$DISTVERSION" ] && misuse DISTVERSION
 
-apt-get install -qq -y gunicorn g++ python3-venv >/dev/null
+apt-get install -qq -y g++ python3-venv >/dev/null
 
 install_nginx
 
@@ -54,6 +54,9 @@ if [ $POSTGIS_TWO = Yes ]
 then
     su -l -c "psql -c 'ALTER USER $UNIX_USER WITH NOSUPERUSER'" postgres
 fi
+
+# Install gunicorn within the virtual environment
+su -l -c "source $DIRECTORY/virtualenv-mapit/bin/activate && pip install gunicorn==21.0.1" "$UNIX_USER"
 
 install_sysvinit_script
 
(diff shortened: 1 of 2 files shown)


mikejamesthompson:
If this is the kind of thing that you'd like to merge into the repo then I'd be happy to also contribute documentation changes as well.

dracos:
Thanks, looks good! Happy to merge with doc changes :)

mikejamesthompson:
> Thanks, looks good! Happy to merge with doc changes :)

I was thinking of [these docs](http://code.mapit.mysociety.org/install/install-script/) but can't find them in this repo and I'm not sure what else I would update. Can you point me in the right direction please 🙏 ?

dracos:
Sure, it looks like they're the `gh-pages` branch of this repo.

mikejamesthompson:
> Sure, it looks like they're the `gh-pages` branch of this repo.

Ok, I'll need to do that in a separate PR.

mikejamesthompson:
> Just looking at this now, and an issue is that gunicorn isn't a necessary component of mapit itself, so I'm not sure we want that in install_requires - because it doesn't. Could that be taken out of there, and put back in site-specific-install.sh - but installed into the virtual env as you suggest, rather than installed with apt?

Yes, fair enough.

I've added a couple of commits that make that change, PTAL.

dracos:
Thanks :)

PR #415, opened 2023-11-22
Install script docs update

mikejamesthompson:
This PR is the docs update to accompany #413.

I have arguably strayed a little beyond my purview here, but I wanted to tidy up those light text on light background code snippets! Feel free to amend or ignore as you see fit.

Now looks like:
Screenshot 2023-11-22 at 21 44 22

You might also like to remove some of the footer links as well (the Pombola one is 404-ing).


dracos:
Thanks!

PR #424, opened 2024-03-13
Add max_generation parameter.

dracos:
Fixes https://github.com/mysociety/societyworks/issues/4162

### DIFF
--- a/mapit/models.py
+++ b/mapit/models.py
@@ -6,6 +6,7 @@ import itertools
 from django.contrib.gis.db import models
 from django.conf import settings
 from django.db import connection
+from django.db.models import Q
 from django.db.models.query import RawQuerySet
 from django.utils.encoding import smart_str
 from django.utils.functional import cached_property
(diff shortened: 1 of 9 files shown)


mapit/models.py
```diff
@@ -47,21 +48,26 @@ def new(self):
...
-        except ValueError:
-            raise ViewException(format, _('Bad min_generation specified'), 400)
-        if not min_generation:
-            min_generation = generation
+        args = {}
+        for q in ('generation', 'min_generation', 'max_generation'):
+            try:
+                args[q] = int(request.GET.get(q, 0))
+            except ValueError:
+                raise ViewException(format, _('Bad generation specified'), 400)
```
davea, line 56: This error would be more helpful if it mentioned which param had the bad value. Maybe something like:
```suggestion
                raise ViewException(format, _('Bad value specified for %s parameter') % q, 400)
```

davea (approved):
One minor suggestion for UX improvement but overall looks good! 👍

PR #430, opened 2024-05-27
fix(mapit_import): Filter by area type when looking for existing areas

### LINKED ISSUE #429: mapit_filter: Unable to filter by type and code
### Problem

When importing using `mapit_import` you can’t import codes while also filtering for an area type:

- If you specify both `--code_field`/`--code_type` and `--area_type_code` then the area type is ignored when searching for existing areas.
- If you specify only `--area_type_code` then the codes are not imported.

When one code is used in multiple area types this can lead to areas of a different type being modified.

### Steps to reproduce

I observed this when importing the WMCF areas because there are a few scottish constituencies which keep their area code. Importing them with:

```bash
python manage.py mapit_import --area_type_code=WMCF --country_code=first-letter --name_type_code=O --code_field=gss_code --code_type=gss parl_constituencies_2025.gpkg
```

Will only create 632 WMCF areas although there are 650 areas in the dataset.  Turns out the culprit is this bit of code in `mapit/add_areas_from_file/core.py`:

```python
            if parameters.new:  # Always want a new area
                raise Area.DoesNotExist
            if code:
                matching_message = "code %s of code type %s" % (
                    code,
                    parameters.code_type,
                )
                areas = Area.objects.filter(
                    codes__code=code, codes__type=parameters.code_type
                ).order_by("-generation_high")
            else:
                matching_message = "name %s of area type %s" % (
                    name,
                    parameters.area_type,
                )
                areas = Area.objects.filter(
                    name=name, type=parameters.area_type
                ).order_by("-generation_high")
```

### Workaround

- On the initial import use `--preserve`  to force creating new areas.
- No workaround for updating areas.

### Proposed change

If both codes and an area type is specified when importing the script should filter by both. In particular it should never touch areas of another type than the one specified as `--area_type_code`.

torotil:
This makes the import command always filter by area type (which is a mandatory argument).

Fixes #429

### DIFF
--- a/mapit/add_areas_from_file/core.py
+++ b/mapit/add_areas_from_file/core.py
@@ -269,7 +269,7 @@ def run(filename, parameters, logger=None):
 
             m = areas[0]
             if logger:
-                logger.debug("    found the area")
+                logger.debug(f"    found the area {m}")
             if parameters.preserve:
                 # Find whether we need to create a new Area:
                 previous_geos_geometry = m.polygons.aggregate(
--- a/mapit/add_areas_from_file/core.py
+++ b/mapit/add_areas_from_file/core.py
@@ -243,22 +243,22 @@ def run(filename, parameters, logger=None):
         try:
             if parameters.new:  # Always want a new area
                 raise Area.DoesNotExist
+            filters = {"type": parameters.area_type}
             if code:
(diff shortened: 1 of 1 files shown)


neprune (approved):
LGTM. Thanks!

PR #432, opened 2024-12-06
Use ST_Covers not ST_Contains for subdivision point lookup

davea:
ST_Contains doesn’t match if the point is on the boundary of the subdivision.


dracos (commented):
It would be good to have a test, the shapes in test_views.py presumably have subdivided entries automatically added that you could find a boundary point on?

dracos (approved):
Thanks :)

PR #435, opened 2025-05-29
Using latest upstream mapit for Sinar/mapit

samqi:
- We are using latest upstream mapit for Sinar Project


samqi:
ops, 
- I may have mistakenly merge Sinar/mapit into main mysociety/mapit 
- I thought I was merging mysociety/mapit latest into Sinar/mapit and not the other way round
- If necessary, someone might want to revert this PR to its original state

PR #444, opened 2026-05-17
Return JSON for 404s, matching the rest of the API

### LINKED ISSUE #421: API returns HTML for 404s
The API documentation says:

> All calls return JSON, and you can generally get an HTML representation by adding .html on the end.

and

> Whenever an error is returned from MapIt, it is as a dictionary with an error key …

but for example https://mapit.mysociety.org/postcode/F-54110 returns a HTML page.

c-tonneslan:
Per #421 and @dracos' comment there, hitting a URL the router doesn't recognise currently falls through to Django's HTML 404 page even though the docs say the API always returns JSON. `handler500` is already wired up to `mapit.shortcuts.json_500`; just mirror the pattern with a `json_404` and register it as `handler404`.

`get_object_or_404` still goes through `ViewException` so endpoint-level 404s are unchanged.

Closes #421.

### DIFF
--- a/mapit/shortcuts.py
+++ b/mapit/shortcuts.py
@@ -101,7 +101,7 @@ def json_500(request):
     return output_json({'error': "Sorry, something's gone wrong."}, code=500)
 
 
-def json_404(request, exception=None):
+def json_404(request, *args, **kwargs):
     return output_json({'error': "Not found."}, code=404)
 
 
--- a/mapit/shortcuts.py
+++ b/mapit/shortcuts.py
@@ -101,6 +101,10 @@ def json_500(request):
     return output_json({'error': "Sorry, something's gone wrong."}, code=500)
 
 
+def json_404(request, exception=None):
+    return output_json({'error': "Not found."}, code=404)
+
+
 def set_timeout(format):
     cursor = connection.cursor()
     timeout = 10000 if format == 'html' else 10000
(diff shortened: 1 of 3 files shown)


dracos:
Thanks for your contribution, but I'm afraid this is too broad - we would still want people coming to the website and going to a page that doesn't exist (for whatever reason) to get a normal HTML 404 page, not a JSON response, which would only be for the API endpoints. Perhaps the solution is to allow more things through the url handler, and then have them dealt with there.

c-tonneslan:
Good point, thanks. I dropped the global `handler404` and added catch-all patterns at the end of `mapit/urls.py` for the API URL prefixes (`postcode`, `area`, `areas`, `point`, `nearest`, `code`, `generations`, `types`). Anything not matching one of those prefixes still falls through to the project-level HTML 404.

Spot-checked with the regexes locally:

- `/postcode/F-54110` (the one from #421) → JSON 404
- `/postcode/SW1A1AA` → unchanged
- `/postcode/` → still the HTML form page
- `/generations.xml` → JSON 404
- `/generations.json` → unchanged
- `/random-path` → HTML 404

Let me know if you'd rather inline the catch-all view instead of keeping `json_404` next to `json_500` in shortcuts.

dracos:
I like the idea of fixing this one, thanks, hope to be able to take a look at it soon.

PR #446, opened 2026-05-18
Narrow bare excepts in import commands to DoesNotExist

c-tonneslan:
The lookups in `get_area_type` / `get_name_type` / `get_country` / `get_code_type` and the area-union geometry filter were all really catching the model's `DoesNotExist`. Saying so explicitly stops a typo or a `KeyboardInterrupt` during the `input()` prompt from getting silently swallowed.


dracos:
`A KeyboardInterrupt during the input() prompt from getting silently swallowed` - this is incorrect, because the input() is already in the except block?

PR #447, opened 2026-05-18
GEOS_JSONEncoder: catch AttributeError, not anything

c-tonneslan:
The encoder's `default()` probes for an `o.json` attribute and falls back to the parent class when it isn't there. A bare except also swallows the parent's `TypeError` for an unencodable type, which is what we'd actually want to surface.

### DIFF
--- a/mapit/shortcuts.py
+++ b/mapit/shortcuts.py
@@ -17,7 +17,7 @@ class GEOS_JSONEncoder(DjangoJSONEncoder):
     def default(self, o):
         try:
             return o.json  # Will therefore support all the GEOS objects
-        except:
+        except AttributeError:
             pass
         return super(GEOS_JSONEncoder, self).default(o)
 


dracos:
I don't understand this. No error from the parent is swallowed, that I can see? Sorry if I've misunderstood.
I see no real issue with having a bare `except` here, so that any possible issue is skipped.
Thanks for these contributions, but this is code that has worked fine for many years. Could you please give a quick introduction; it feels a bit odd to suddenly spray many of our repositories with many small fixes, which I'm sure are good in general, but not really necessary and come up against a lot of competing priorities as a very small team.

c-tonneslan:
My apologies. I'm a working dev trying to break into open source while job hunting and got carried away after enjoying reading through mapit. Closing this and the rest as well. Thanks for your patience.

PR #449, opened 2026-09-10
[UK] Allow CED GSS codes to be assigned from ONS geopackage rather than CSV file

davea:
ONS don't seem to publish a CSV with the required fields as often as we'd like, but they do provide a GeoPackage with the same data in - switch to matching MapIt's CED Areas spatially and assiging GSS codes that way instead.

Removes the need for all the name overrides too.


dracos (commented):
Looks good. Do we want to have a check for if the area already has a GSS code, that it matches? If we ran this in the past. (But I can't see any CEDs with a code in the db, so perhaps we didn't?)

davea:
I moved from `area.codes.get_or_create` to `area.codes.update_or_create` for that reason - there was at least one (I forget which now) that appeared to have changed GSS code.

dracos (approved):
Code looks fine; but a GSS code should never change, so that's a bit odd - probably worth checking which it was in case there's an issue for some reason. (I guess perhaps there was a boundary change but we didn't pick it up properly?)

